PR #142, opened 2014-11-06
Add upsert support to `Model.save`, `Document.save`, and `Document.saveAll`

mstade:
This adds a second (optional) parameter to `Model.save`, `Document.save`,
and `Document.saveAll`; which allows you to specify how conflicts should be
handled when inserting documents.

When `options.conflict` is set to either `update` or `replace`, documents
will be updated or replaced when inserted if they already exist in the
database. If the value is set to anything else it will revert to `error`.

The default conflict resolution is `error`, which is in line with how it
worked before since that's what RethinkDB defaults to.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -747,26 +747,32 @@ Document.prototype.__validate = function(doc, schema, prefix, options, originalD
     }
 }
 
-Document.prototype.save = function(callback) {
+Document.prototype.save = function(options, callback) {
     var self = this;
 
-    return self._save({}, false, {}, callback);
+    if (typeof options === 'function') {
+        callback = options;
+        options = { conflict: 'error' };
+    }
+
+    return self._save({}, false, {}, options, callback);
 }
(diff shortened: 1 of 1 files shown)


neumino:
Err I'm really sorry, I missed this pull request -_-

I'll add options for `Model.save` but I'm a bit more worried about `Document.save` because of the joined documents. I'll try to merge this, but since I've refactored `Document.save`, this seems hard.

Again, really sorry @mstade for missing your pull request :/

xpepermint:
+1

mstade:
@neumino Don't worry about it buddy, it's easily done in all the noise! Since publishing this PR I've moved on to other things and am no longer actively using either RethinkDB and thus not thinky either. At the time, missing upsert was a big deal(tm) and a pretty big gotcha. Whether this is still true with current versions of thinky I wouldn't know, so you may want to consider just closing this PR or creating a new issue to deal with this for more the up-to-date code base.

Finally – thanks for your great efforts to bring better tools to the developer community!

PR #171, opened 2015-01-16
Do not validate reql terms in partial update

simonratner:
Match behaviour of array and object fields to other types.


neumino:
Looks good to me.
Thanks @simonratner!

PR #173, opened 2015-01-21
Fix purge on models referenced by a hasMany join

simonratner:
Seems to have been broken by the hook refactor in aa748ff4e00101859847c23e4ccfa74b4e60ddbe.


neumino:
Arg sorry, I missed the PR in my mails. Looks good to me. Thanks again @simonratner

PR #195, opened 2015-03-08
Add .id(), .optional() and .required() as aliases

### LINKED ISSUE #185: enforce_missing: true option should allow id to be missing
I'm using Thinky with the global ``enforce_missing: true`` option.

In addition, I've defined several objects in Thinky with string ids like so:

    var Class = thinky.createModel('Class', {
      id:       type.string(),
      name:     type.string(),
      schoolId: type.string()
    });

When attempting to save a new object, I get an error that ``id`` is not set, which seems odd, since ``id`` is about to get created by Rethink.

I've worked around it by changing the ``id`` definition to the following:

          id:       type.string().options({enforce_missing: false}),

It would be nice to not have to do this all the time, as it's quite verbose.  I can think of two reasonable options:

1. Have Thinky treat **``id``** as a special name and allow it to be missing even when the global ``enforce_missing`` flag is ``true``.

2. Add a new function, ``type.id()``, that is an alias for ``type.string()``, but also sets the ``enforce_missing`` flag to ``false``.

Then I could instead write:

    var Class = thinky.createModel('Class', {
      id:       type.id(),
      name:     type.string(),
      schoolId: type.string()
    });

mindjuice:
Fixes #185.
Added .id() as an alias for string().optional().
Added .optional() as an alias for .options(enforce_missing: false}).
Added .required() as an alias for .options(enforce_missing: true}).
Modified .options() to merge the given options into the existing options to allow you to write something like: ``myProp: type.string().optional().options({enforce_type: "strict"})``
Added some simple tests for mergeOptions().

### DIFF
--- a/lib/type/array.js
+++ b/lib/type/array.js
@@ -27,6 +27,21 @@ TypeArray.prototype.required = function() {
 }
 
 
+TypeArray.prototype.noTypeCheck = function() {
+  return this.options({enforce_type: 'none'});
+}
+
+
+TypeArray.prototype.looseTypeCheck = function() {
+  return this.options({enforce_type: 'loose'});
+}
+
+
+TypeArray.prototype.strictTypeCheck = function() {
+  return this.options({enforce_type: 'strict'});
+}
+
+
 TypeArray.prototype.min = function(min) {
   if (min < 0) {
     throw new Error("The value for `min` must be a positive integer");
--- a/lib/type/array.js
+++ b/lib/type/array.js
@@ -27,18 +27,13 @@ TypeArray.prototype.required = function() {
 }
 
 
-TypeArray.prototype.noTypeCheck = function() {
-  return this.options({enforce_type: 'none'});
-}
-
-
-TypeArray.prototype.looseTypeCheck = function() {
-  return this.options({enforce_type: 'loose'});
-}
-
-
-TypeArray.prototype.strictTypeCheck = function() {
-  return this.options({enforce_type: 'strict'});
+TypeArray.prototype.checkType = function(isChecked, allowNull) {
+  allowNull = typeof allowNull !== 'undefined' ? allowNull : true;
+  var value = 'none';
+  if (isChecked) {
+    value = allowNull ? 'loose' : 'strict';
+  }
+  return this.options({enforce_type: value});
 }
 
 
--- a/lib/type/array.js
+++ b/lib/type/array.js
@@ -27,16 +27,6 @@ TypeArray.prototype.required = function() {
 }
 
 
-TypeArray.prototype.checkType = function(isChecked, allowNull) {
-  allowNull = typeof allowNull !== 'undefined' ? allowNull : true;
-  var value = 'none';
-  if (isChecked) {
-    value = allowNull ? 'loose' : 'strict';
-  }
-  return this.options({enforce_type: value});
-}
-
-
 TypeArray.prototype.min = function(min) {
   if (min < 0) {
     throw new Error("The value for `min` must be a positive integer");
--- a/lib/type/array.js
+++ b/lib/type/array.js
@@ -12,8 +12,18 @@ function TypeArray() {
 
 
 TypeArray.prototype.options = function(options) {
-  this._options = options;
+  this._options = util.mergeOptions(this._options, options);
   return this;
+}
+
+
+TypeArray.prototype.optional = function() {
+  return this.options({enforce_missing: false});
+}
+
+
+TypeArray.prototype.required = function() {
+  return this.options({enforce_missing: true});
 }
 
 
--- a/lib/type/boolean.js
+++ b/lib/type/boolean.js
@@ -22,18 +22,13 @@ TypeBoolean.prototype.required = function() {
 }
 
 
-TypeBoolean.prototype.noTypeCheck = function() {
-  return this.options({enforce_type: 'none'});
-}
-
-
-TypeBoolean.prototype.looseTypeCheck = function() {
-  return this.options({enforce_type: 'loose'});
-}
-
-
-TypeBoolean.prototype.strictTypeCheck = function() {
-  return this.options({enforce_type: 'strict'});
+TypeBoolean.prototype.checkType = function(isChecked, allowNull) {
+  allowNull = typeof allowNull !== 'undefined' ? allowNull : true;
+  var value = 'none';
+  if (isChecked) {
+    value = allowNull ? 'loose' : 'strict';
+  }
+  return this.options({enforce_type: value});
 }
 
 
--- a/lib/type/boolean.js
+++ b/lib/type/boolean.js
@@ -22,6 +22,21 @@ TypeBoolean.prototype.required = function() {
 }
 
 
+TypeBoolean.prototype.noTypeCheck = function() {
+  return this.options({enforce_type: 'none'});
+}
+
+
+TypeBoolean.prototype.looseTypeCheck = function() {
+  return this.options({enforce_type: 'loose'});
+}
+
+
+TypeBoolean.prototype.strictTypeCheck = function() {
+  return this.options({enforce_type: 'strict'});
+}
+
+
 TypeBoolean.prototype.default = function(fnOrValue) {
   this._default = fnOrValue;
   return this;
--- a/lib/type/boolean.js
+++ b/lib/type/boolean.js
@@ -7,8 +7,18 @@ function TypeBoolean() {
 
 
 TypeBoolean.prototype.options = function(options) {
-  this._options = options;
+  this._options = util.mergeOptions(this._options, options);
   return this;
+}
+
+
+TypeBoolean.prototype.optional = function() {
+  return this.options({enforce_missing: false});
+}
+
+
+TypeBoolean.prototype.required = function() {
+  return this.options({enforce_missing: true});
 }
 
 
--- a/lib/type/boolean.js
+++ b/lib/type/boolean.js
@@ -22,16 +22,6 @@ TypeBoolean.prototype.required = function() {
 }
 
 
-TypeBoolean.prototype.checkType = function(isChecked, allowNull) {
-  allowNull = typeof allowNull !== 'undefined' ? allowNull : true;
-  var value = 'none';
-  if (isChecked) {
-    value = allowNull ? 'loose' : 'strict';
-  }
-  return this.options({enforce_type: value});
-}
-
-
 TypeBoolean.prototype.default = function(fnOrValue) {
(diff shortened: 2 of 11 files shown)


mindjuice:
Let me know if this is okay.  I can make any changes or additions as necessary.

mindjuice:
Derp!  I just realized a few things wrong here:

1. ``enforce_extra`` isn't a boolean type as I wrongly assumed.  Can you confirm my understanding of the options?
  - ``strict`` - Extra fields are not allowed and will cause validation to fail
  - ``remove`` - Extra fields are not allowed, but will simply be removed instead of failing validation
  - ``none`` - Extra fields are allowed

2. You mentioned possibly having ``type.string().removeExtra()``, which I forgot about.  Thinking about this though, I don't think ``removeExtra()`` makes sense on an individual field, does it?  Having ``removeExtra()`` on the Schema itself makes sense though, so I can add that.  Suggestions on what the names should be?  Maybe ``removeExtraProps()``, ``failExtraProps()``, ``allowExtraProps()``?  Or just  ``removeExtra()``, ``failExtra()``, ``allowExtra()``?

3. We should do something similar for ``enforce_type``.  Maybe ``strictTypeCheck()``, ``looseTypeCheck()``, ``noTypeCheck()``?  Or ``strictTyping()``, ``looseTyping()``, ``noTyping()``?

4. Do you think we need unit tests for these aliases?  They are pretty simple, but I suppose they could still be wrong.

mindjuice:
With regard to item #2 above, ``removeExtra / failExtra / allowExtra`` does make sense if applied to an ``object`` type.  The setting overrides the schema setting, if any.  I think that's the way it works now too.

mindjuice:
I've updated the PR with most of the changes I mentioned above.  I haven't added any additional unit testing yet (except for the mergeOptions() function from the initial version of the PR).

neumino:
Thanks a lot @mindjuice! I look ed a the code, and it looks good to me.

I'm also sorry for the late answer, I spent quite some time thinking about the names to use. I actually started to implement `optional` and `required` at some point, but then stop on how to call the remaining methods.

I'm not sure if these are good suggestions, especially since I'm not a native English speaker, but here they are:
- `removeExtra` and `allowExtra` sound good to me.
- I'm not a big fan of `failExtra`, does `forbidExtra` sound better? Maybe we could just have `allowExtra` to take a boolean. I have a slight preference for that as people don't have to remember one more method name (it's also what https://github.com/hapijs/joi is doing with `allowUnknown` for example).
- I would also rename `looseTypeCheck` to `allowNull` as it's easier to understand. And rename `strictTypeCheck` to `forbidNull` or have `allowNull` to take a boolean as argument.
- `noTypeCheck` could be rename to `allowAll` maybe? It feels better when reading the chained methods.

TL;DR: We would have
- `removeExtra()`
- `allowExtra(<boolean>)`
- `allowNull(<boolean>)`
- `allowAll()`

What do you think @mindjuice?

mindjuice:
Thanks for the feedback.  Your new names sound good.  I'll try to update this tonight.

neumino:
Awesome, thanks @mindjuice :)

mindjuice:
I've updated my code for the ``removeExtra/allowExtra`` methods as you suggested, however, looking at the names for the type-related functions again, the one thing that seems a bit off is that there is no mention of 'type' for ``allowNull()`` and ``allowAll()``, yet they are related to the ``enforce_type`` flag.

Here is a summary of the current version for the ``enforce_extra`` flag:
* **``allowExtra(allowed)``** - If ``allowed`` is ``true``, then all extra properties are allowed (corresponds to  ``enforce_extra = 'none'``).  If ``allowed`` is ``false``, then validation will fail if any additional properties are present (corresponds to  ``enforce_extra = 'strict'``).
* **``removeExtra()``** - If any extra properties are present, they will be removed during validation (corresponds to  ``enforce_extra = 'remove'``).

And what I have implemented for the ``enforce_type`` flag is a single method with two parameters:
* **``checkType(isChecked, allowNull)``** - Turns on/off type checking for this property.  If ``isChecked`` is ``false``, then no type checking is done (corresponds to ``enforce_type`` = 'none').  If ``isChecked`` is ``true``, then the value must be of the type specified in the Model.  Additionally, when ``isChecked`` is ``true``, if ``allowNull`` is ``true``, then the property can also be ``null`` (corresponds to ``enforce_type`` = 'loose').  If ``allowNull`` is ``false``, then ``null`` is not allowed (corresponds to ``enforce_type`` = 'strict').

The ``checkType()`` method still feel a little awkward still though, since it conflates two different concepts: 1) whether to check the type and 2) whether or not it can be null).  If we had two boolean separate flags (e.g., ``enforce_type`` and ``allow_null``), then we could define directly corresponding methods:

* **``checkType(isChecked)``** - Turns type checking on or off.  If ``true``, then the type must be of that specified in the Model.  In this case, the value of ``allowNull()`` is also checked to see if ``null`` values are are acceptable.  If ``false``, then any value is acceptable and the value of ``allowNull()`` does not matter.
* **``allowNull(allowed)``** - If ``true``, then ``null`` values are allowed.  If ``false``, then they will cause a validation error.

This would require some refactoring of the ``enforce_type`` option though.

Let me know what you think.

neumino:
I think it's nicer to split `checkType` in two method. It's a little of refactoring, but it shouldn't be too hard.
I'm out of town for the next two days, but I can do that if you don't have time @mindjuice.

I'm not a big fan of the name `checkType(isChecked)` too, though I have to admin `allowAll(<boolean>)` is not really better. We could have `any(allowAny)` maybe?

mindjuice:
If I understand this properly (please correct me if I'm wrong), the following would be the way to use the ``options()`` method with the  ``enforce_type`` flag.

    titi: type.boolean().options({enforce_type: "none"}),  // Can be any value, including null
    toto: type.string().options({enforce_type: "loose"})   // Can only be string or null

It seems a bit odd to say "This thing is a string, but accept any type."

What if instead of having ``allowAny()/allowAll()``, we add a ``TypeAny`` type, which has no validation performed on it?

    /**
     * Create a new TypeAny object.
     * @return {TypeAny}
     */
    Type.prototype.any = function() {
      return new TypeAny();
    }

Then you can write:

    titi: type.any().allowNull(true),      // Can be any value, including null
    toto: type.string().allowNull(true),   // Can only be string or null
    tata: type.any(),                      // Can be any value, but cannot be null
    tutu: type.string()                   // Can only be string, and cannot be null

Compared to my current code, the changes would be:

* Remove ``checkType()`` throughout
* Create ``TypeAny``
* Add ``allowNull(allow)`` to all types (e.g., ``TypeString``, ``TypeInteger``)

This lets you clearly express your intention to allow a given field to be of any type in the schema and separates the concept of type checking from null checking.

mindjuice:
Hi @neumino, I've temporarily removed the ``checkType()`` methods so that we can get the other changes merged sooner.  If that's OK with you, I can create a separate issue for tracking the type checking changes discussed above.

neumino:
@mindjuice -- `type.any()` sounds good to me (and better than what I suggested earlier.

I would have `type.any()` allow any type including null.
I would also have `TypeAny` implement all the methods from other types but do nothing. The reason is that if people have a bug with a field, they can momentarily replace something like
```
{
    // ... other fields
    email: type.string().email().allowNull(false)
}
``` 
with
```
{
    // ... other fields
    email: type.any().email().allowNull(false)
}
``` 

One last thing, we should keep the same default as now (that is to say `null` is allowed).

Sorry for the slow responses, I was out of town. I looked at your changes and things looks good to me.
I'll try to merge that tomorrow (or this week end at worse) as there are also some docs to take care of, though that should be pretty easy to do.

neumino:
Setting up for `type.any()` and `allowNull` and `required`/`optional`.
Doing that now based on this PR.

neumino:
Ok, this should be in branch `next` with https://github.com/neumino/thinky/pull/198
I'll write some docs and release it tomorrow or this week end.

Thanks @mindjuice :)

mindjuice:
Cool!  Thanks @neumino.  I also started the work in type.any(), but didn't finish.  I'll leave it to you then. :)

Below is any.js that I started on in case there is anything in there that is helpful.  I left ``_options`` and ``_default`` to allow for those even with any type.

The only other change I had time to make was in ``type/index.js`` to add the .any() function.

    var util =       require(__dirname+'/../util.js');
    var validator =  require("validator");


    /**
     * Create a new TypeAny object
     */
    function TypeAny() {
      this._options = undefined;
      this._default = undefined;
    }


    TypeAny.prototype.options = function(options) {
      this._options = util.mergeOptions(this._options, options);
      return this;
    }


    TypeAny.prototype.optional = function() {
      return this.options({enforce_missing: false});
    }


    TypeAny.prototype.required = function() {
      return this.options({enforce_missing: true});
    }


    TypeAny.prototype.default = function(fnOrValue) {
      this._default = fnOrValue;
      return this;
    }


    TypeAny.prototype.validate = function(obj, prefix, options) {
      // The only requirement is that this function not throw an exception
    }


    /**
     * Look for a default value or default function, and append an object to `defaultFields`.
     * @param {string} prefix The prefix leading to `str`.
     * @param {Array.<Object>} defaultFields The default fields to generate
     * @param {Array.<Object>} virtualFields The virtual fields to generate
     * @return {TypeAny}
     */
    TypeAny.prototype._getDefaultFields = function(prefix, defaultFields, virtualFields) {
      if (this._default !== undefined) {
        defaultFields.push({
          path: prefix,
          value: this._default,
        });
      }
      return this;
    }


    /**
     * Pass-through chaining function that just returns the object.  Used by the functions
     * declared below.
     */
    TypeAny.prototype._chain = function() {
      return this;
    }


    /**
     * Pass-though functions to allow a user to temporarily replace a specific type with any.
     *
     * e.g., replace this:   email: type.string().email().allowNull(false)
     *       with this:      email: type.any().email().allowNull(false)
     */
    TypeAny.prototype.min = _chain;
    TypeAny.prototype.max = _chain;
    TypeAny.prototype.integer = _chain;
    TypeAny.prototype.length = _chain;
    TypeAny.prototype.regex = _chain;
    TypeAny.prototype.alphanum = _chain;
    TypeAny.prototype.email = _chain;
    TypeAny.prototype.lowercase = _chain;
    TypeAny.prototype.uppercase = _chain;
    TypeAny.prototype.schema = _chain;
    TypeAny.prototype.setKey = _chain;
    TypeAny.prototype.default = _chain;
    TypeAny.prototype.validator = _chain;
    TypeAny.prototype.enum = _chain;


    module.exports = TypeAny;

neumino:
Released in 1.16.7, thanks again @mindjuice :)

PR #198, opened 2015-03-16
New Error Type: ValidationError

primitive-type:
It would be helpful to have a new error type for validation errors, similar to the ```DocumentNotFound``` error type defined in https://github.com/neumino/thinky/blob/master/lib/errors.js

The use case I am imagining is for HTTP status code responses in a RESTful API. You could more granularly report to the API consumer whether an error was their fault due to submission of bad data which the thinky model rejects (400), a particular document not being found (404 - which the ```DocumentNotFound``` type is already helpful for), or some other failure (500). 

I believe this new error could be thrown instead of the more generic Error type in https://github.com/neumino/thinky/blob/master/lib/schema.js

If you think this makes sense, I would be happy to try my hand at a pull request for this.

Thanks!

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -3,6 +3,7 @@ var type =          require(__dirname+'/type/index.js');
 var util =          require(__dirname+'/util.js');
 var Promise =       require('bluebird');
 var EventEmitter =  require('events').EventEmitter;
+var Errors =        require(__dirname+'/errors.js');
 
 
 /**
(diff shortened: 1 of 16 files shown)


primitive-type:
I think I caught all instances where a `ValidationError` should be returned instead of a generic one. Let me know if there was anything I missed.

neumino:
Ok, this should be in branch `next` with https://github.com/neumino/thinky/pull/195
I'll write some docs and release it tomorrow or this week end.

Thanks @primitive-type :)

primitive-type:
@neumino Great. Thanks for thinky! It has been a real joy to use.

neumino:
Released in 1.16.7, thanks again @primitive-type :)

PR #205, opened 2015-03-31
Fix default() doc

francois2metz:


neumino:
Looks good, thanks @francois2metz!

PR #222, opened 2015-04-22
Fix promise helpers to return promises

simonratner:
These currently return an instance of Query rather than correctly executing the query and returning the promise with attached handler.

### DIFF
--- a/test/query.js
+++ b/test/query.js
@@ -6,6 +6,9 @@ var Document = require(__dirname+'/../lib/document.js');
 var util = require(__dirname+'/util.js');
 var assert = require('assert');
 var Promise = require('bluebird');
+Promise.onPossiblyUnhandledRejection(function() {
+    // suppress possibly unhandled rejection warnings
+});
 
 var modelNameSet = {};
 modelNameSet[util.s8()] = true;
@@ -978,7 +981,6 @@ describe('then', function() {
     User.then(function() {
         done();
     });
-
   });
   it('should run the query', function(done) {
     var name = util.s8();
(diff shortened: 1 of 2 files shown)


simonratner:
See also related neumino/rethinkdbdash/pull/83

simonratner:
Any chance of merging this? I would like to remove the monkey-patch in my code :)

neumino:
Sorry I completely missed this. Looking into it.

test/query.js
```diff
@@ -6,6 +6,9 @@ var Document = require(__dirname+'/../lib/document.js');
 var util = require(__dirname+'/util.js');
 var assert = require('assert');
 var Promise = require('bluebird');
+Promise.onPossiblyUnhandledRejection(function() {
```
neumino: Can you remove this and properly catch errors?
While it may be justified, it may catch other errors that we don't want to.

test/query.js
```diff
@@ -990,7 +992,104 @@ describe('then', function() {
...
+  });
+  it('should return a promise', function(done) {
+    var name = util.s8();
+
+    var Query = thinky.Query;
+    var r = thinky.r;
+    var User = thinky.createModel(modelNames[0], {id: String}, {init: false});
+
+    User.filter({}).then(function() {
+        // no-op
```
neumino: Why don't you call `done` here instead of chaining `spread`?

```
User.filter({}).then(function(result) {
  assert(Array.isArray(result));
  done();
}).catch(done)
```

neumino: Same questions for the other tests

neumino:
I'm fine with the modification in lib/query.js - and it's a good one, again sorry to have missed it :)

I'm a bit confused on the tests. Why do you need `spread`? And `Promise.onPossiblyUnhandledRejection`

simonratner:
Thanks for taking a look. The use of `spread` was to ensure that what is returned is actually a promise rather than a `Query` (which still had all those methods, just broken, hence the one existing test was passing and not catching the bug). `Promise.onPossiblyUnhandledRejection` is there for the same reason, one of the tests has an uncaught rejection intentionally as a passing condition, but that results in scary console spam from `bluebird`. I should've contained `Promise.onPossiblyUnhandledRejection` to just that test though.

I agree this is a pretty hacky way to do it, it would be better to check the type of the object returned. I'll take another pass at these tests today and see if I can get something better (for some reason, I couldn't get it to work right the first time I tried).

test/query.js
simonratner: To make sure that `done` is called from a completed promise chain, rather than just called by `Query` as part of constructing the term.

simonratner:
Updated with cleaner tests, not sure what I was smoking before (._.)

neumino:
Looks good to me, thanks @simonratner :)

PR #229, opened 2015-05-07
Refactoring and cleanup of Query's _execute() function

rasapetter:
Here are some proposed changes whose purpose it is to improve readability on the rather complex _execute() function in query.js. There were some low hanging fruit in the form of redundant if/else statements and unused variables, however, most of the branch deals with breaking up the function into separate parts.

I'm fairly new to Thinky (and RethinkDB for that matter) so please forgive me if I've made any rookie mistakes.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -210,24 +210,32 @@ Query.prototype._execute = function(options, parse, callback) {
               }
             }
           })
+          return;
         }
-        else if ((result != null) && (typeof result.getType === 'function') &&
-            ((result.getType() === 'Feed') ||
-            (result.getType() === 'OrderByLimitFeed') ||
-            (result.getType() === 'UnionedFeed'))) {
-          var feed = new Feed(result, self._model);
-          resolve(feed);
+
+        if (result != null && typeof result.getType === 'function') {
+          var resultType = result.getType();
+          if (resultType === 'Feed' ||
+            resultType === 'OrderByLimitFeed' ||
+            resultType === 'UnionedFeed'
+          ) {
+            var feed = new Feed(result, self._model);
+            resolve(feed);
+            return;
+          }
+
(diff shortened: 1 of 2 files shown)


rasapetter:
Whoops! Errors. Patch coming up...

rasapetter:
Yup, that's better. It took a while to sort things out due to a misunderstanding early in the branch. Seems OK now - all tests pass and the code should be much easier to read, IMO.

neumino:
Gosh, I'm really sorry @rasapetter, I just found out your pull request while merging #222 -_-

I quickly looked at the changes, they look good to me, and it's actually something I thought about doing, so thanks!
I'll take a more careful look tomorrow or this week end and should merge it.

Thanks for the cleaning :)

rasapetter:
Did you run into any issues here? Or is, perhaps, time an issue? :)

neumino:
Sorry, I got too focused on reqlite. Just checked a few more things, everything looks good.

Thanks a lot @rasapetter for the clean up, rI really appreciate it :)

primitive-type:
Hi @neumino and @rasapetter,

I just wanted to make note of something I came across today. In my codebase, previously, I was calling the ```delete``` method on my models in some tests, as well as using the ```count``` method on a result set (without using ```execute```). After updating thinky to the latest version today, both of these cases started throwing thinky validation errors (correctly, I think) until I added the ```execute``` method. 

In other words, not sure if you were already aware, but this PR seems to have improved the strictness of ```run``` vs. ```execute```. At least, that is my best guess as to what happened since these parts of my code haven't changed in weeks.

Thanks!

neumino:
It is actually not related to this PR, but to https://github.com/neumino/thinky/pull/222
And the `npm publish` command somehow didn't release 2.0.4

PR #231, opened 2015-05-11
refactor(errors): standardize errors and clean up definitions

mbroadst:
- All errors now end with Error, and follow node Error convention
  - Simplified error inheritance with a base error class
  - Added ValidationErrorItem, providing a path for the future for
    multiple validation errors contained in a single error

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,39 +1,73 @@
+'use strict';
+
+var util = require('util');
+
+/**
+ * The base class for all thinky errors
+ *
+ * @constructor
+ * @alias Error
+ */
+function ThinkyError(message) {
+  this.name = 'ThinkyError';
+  this.message = message || 'ThinkyError';
+  Error.captureStackTrace && Error.captureStackTrace(this, this.constructor);
+};
+util.inherits(ThinkyError, Error);
+
 /**
  * DocumentNotFound error which is returned when `get` returns `null`.
  * "Extends" Error
  */
 function DocumentNotFound(message) {
-  Error.captureStackTrace(this, DocumentNotFound);
+  ThinkyError.call(this, message);
+  this.name = "DocumentNotFound";
   this.message = message || "The query did not find a document and returned null.";
 };
-DocumentNotFound.prototype = new Error();
-DocumentNotFound.prototype.name = "Document not found";
-module.exports.DocumentNotFound = DocumentNotFound;
-
+util.inherits(DocumentNotFound, ThinkyError);
 
 /**
  * InvalidWrite error which is returned when an inplace update/replace return
  * anon valid document.
  * "Extends" Error
  */
-function InvalidWrite(message, raw) {
-  Error.captureStackTrace(this, InvalidWrite);
-  this.message = message;
+function InvalidWriteError(message, raw) {
+  ThinkyError.call(this, message);
+  this.name = "InvalidWriteError";
   this.raw = raw;
 };
-InvalidWrite.prototype = new Error();
-InvalidWrite.prototype.name = "Invalid write";
-module.exports.InvalidWrite = InvalidWrite;
-
+util.inherits(InvalidWriteError, ThinkyError);
 
 /**
  * ValidationError error which is returned when validation of a document fails.
  * "Extends" Error
  */
-function ValidationError(message) {
-    Error.captureStackTrace(this, ValidationError);
-    this.message = message;
+function ValidationError(message, errors) {
+  ThinkyError.call(this, message || 'Validation Error');
+  this.name = "ValidationError";
+  this.errors = errors || [];
 };
-ValidationError.prototype = new Error();
-ValidationError.prototype.name = "Document failed validation";
-module.exports.ValidationError = ValidationError;
+util.inherits(ValidationError, ThinkyError);
+
+/**
+ * Validation Error Item
+ * Instances of this class are included in the `ValidationError.errors` property.
+ *
+ * @param {string} field    The field that triggered the validation error
+ * @param {string} value    The value that generated the error
+ * @param {string} message  An error message
+ * @constructor
+ */
+function ValidationErrorItem(field, value, message) {
+  this.field = field || null;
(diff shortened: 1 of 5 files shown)


lib/errors.js
```diff
@@ -1,39 +1,73 @@
...
+
+var util = require('util');
+
+/**
+ * The base class for all thinky errors
+ *
+ * @constructor
+ * @alias Error
+ */
+var ThinkyError = function(message) {
```
neumino: Why the function expression and not the function declaration?

lib/errors.js
```diff
@@ -1,39 +1,73 @@
...
+
 /**
  * DocumentNotFound error which is returned when `get` returns `null`.
  * "Extends" Error
  */
-function DocumentNotFound(message) {
-  Error.captureStackTrace(this, DocumentNotFound);
+var DocumentNotFoundError = function(message) {
+  ThinkyError.call(this, message);
+  this.name = "DocumentNotFoundError";
```
neumino: Why did you move it in the constructor? It's technically not muting, so there's no need to recreate the name every time.

lib/errors.js
```diff
@@ -1,39 +1,73 @@
...
+};
+util.inherits(ThinkyError, Error);
+
 /**
  * DocumentNotFound error which is returned when `get` returns `null`.
  * "Extends" Error
  */
-function DocumentNotFound(message) {
-  Error.captureStackTrace(this, DocumentNotFound);
+var DocumentNotFoundError = function(message) {
```
neumino: Hum, people currently access this error with `thinky.Errors.DocumentNotFound`. While I'm not really opposed to `thinky.Errors.DocumentNotFound`, there is not a reason big enough to break compatibility here.
Could you switch it back to ``DocumentNotFound`?

lib/errors.js
neumino: Same thing for other properties.

lib/errors.js
```diff
@@ -1,39 +1,73 @@
+'use strict';
```
neumino: I'm pretty sure some code doesn't pass if we were to have "use strict" everywhere.

From a more general point of view, I'm still not convinced that "use strict" is that good. It may help to catch a few things, but they are not hard to catch without, and overall it's takes more time than it save.

mbroadst: the only place ```'use strict';``` doesn't work currently is in the schemautils (lib/schema.js), and that's because of a recursive this reference iirc. 

I'm not sure I agree with you about ```'use strict';```, how does it take more time than it saves? Frankly it was basically as easy as adding it to the top of each file, and have that locked in and enforced by jshint.. took no time at all, and you get these benefits from it: 

"Strict mode makes several changes to normal JavaScript semantics. First, strict mode eliminates some JavaScript silent errors by changing them to throw errors. Second, strict mode fixes mistakes that make it difficult for JavaScript engines to perform optimizations: strict mode code can sometimes be made to run faster than identical code that's not strict mode. Third, strict mode prohibits some syntax likely to be defined in future versions of ECMAScript." (from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode))

mbroadst: I'd also like the note that I tested this code very aggressively against the unit tests, and you can see yourself from the wercker run that all of this code *does* have strict mode enabled almost everywhere and passes all tests fine.

lib/errors.js
mbroadst: so it can be easily reused and exported

lib/errors.js
mbroadst: okie doke

lib/errors.js
mbroadst: I moved it all to the constructor because I felt like it made the code more readable to move all manipulation to the same place. Definitely properties like e.g. message need to be computed on a case-by-case basis. Your argument is that it wastes resources or something to not define name as part of the prototype? I'm not sure it's warranted considering the number of potential error objects created in the lifetime of using this module, so I opt for readability in that case

lib/errors.js
marshall007: I think @neumino asking why not continue using function declarations (i.e. `function ThinkyError(message) { ... }`) instead? Either way `ThinkyError` will be defined in this scope so I'm not sure what you mean by easily reused/exported.

mbroadst: oh I thought he was asking about why it was being assigned to the variable. Is there a benefit to the alternative approach (seems there's always more for me to learn about javascript :smile:)

marshall007: @mbroadst see [function declaration hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#Function_declaration_hoisting) (MDN docs), that's basically the only difference.

neumino:
Sorry for the delay, I had a bug in prod yesterday and it took all night to track it down.

I thought the point of your PR was to support returning all errors during a validation instead of the first one. This pull request just create a new "super" class ThinkyError. It's technically a breaking change since `e.constructor` doesn't return the same thing.

I also don' t understand what's the point of introducing a super class. Thinky never throws error, just reject promises. So the only time you would have to check whether an error is from thinky or not is you use something like `Promises.all` with one thinky promise and something else.

mbroadst:
@neumino the point of having a base class for all errors is to isolate shared behaviors between all subclasses: specifically in this case ```Error.captureStackTrace && Error.captureStackTrace(this, this.constructor);``` and giving it a default message if one was accidentally not provided. This is a common pattern throughout most node projects, and I'm not sure I would consider it a breaking change, but we may differ in opinion here (who is really comparing constructors of thrown Errors? I think people are generally catching errors and printing/forwarding their messages, and/or using typeof in which case it would still work fine).

I began working on returning all validation errors within the top level ValidationError, but I figured that while I working in parts of code where I saw that improvements could be made I would refactor as I went (I tend to work this way). Thinky does indeed throw errors when it comes to validation. Generally, I think it's "best practice" to make errors as standardized, and informative as possible, and that's what I was going for here.

PR #232, opened 2015-05-11
general code cleanup

mbroadst:
Apologies for the massive PR, but I hope that I've broken the commits into small enough pieces to make it clear what I actually did. The gist of it is that I used a combination of the closure-linter (fixjsstyle in particular) as well as jshint to apply a first-pass cleanup to the entire codebase (excluding tests for the moment). Additionally, I added a Makefile grouping a number of tasks that otherwise would have gone into the package.json (the Makefile gives a little more power over e.g. select selection). Finally, as part of the cleanup for certain jshint bugs I added some convenience helpers and unrolled some loops.

The tests all still pass, so as far as I can tell I haven't introduced any errors (given your respectable 82% code coverage!)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,35 @@
+{
+  "bitwise":false,
+  "boss":true,
+  "expr":true,
+  "camelcase":false,
+  "curly":false,
+  "eqeqeq":true,
+  "freeze":true,
+  "immed":true,
+  "indent":2,
+  "latedef":"nofunc",
+  "laxbreak":true,
+  "laxcomma":true,
+  "newcap":true,
+  "noarg":true,
+  "node":true,
+  "strict": true,
+  "trailing":true,
+  "undef":true,
+  "esnext":true,
+  "unused":"vars",
+  "nonbsp": true,
+  "maxdepth": 8,
+  "globals": {
+    "Promise": true
+  },
+  "predef": [
+    "describe",
+    "it",
+    "before",
+    "beforeEach",
+    "after",
+    "afterEach"
+  ]
+}
--- a/.jshintrc
+++ b/.jshintrc
@@ -21,6 +21,7 @@
   "unused":"vars",
   "nonbsp": true,
   "maxdepth": 8,
+  "proto": true,
   "globals": {
     "Promise": true
   },
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+REPORTER ?= spec
+TESTS = $(shell find ./test/* -name "*.js")
+NPM_BIN = ./node_modules/.bin
+
+jshint:
+	$(NPM_BIN)/jshint lib
+
+fixjsstyle:
+	fixjsstyle -r lib --strict --jslint_error=all
+
+coverage:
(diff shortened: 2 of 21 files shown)


neumino:
Hum, I'm not sure what to do with this pull requests:
- On one hand there's some good cleaning
- On another, I'm not a big fan of a few changes, typically spacing in the `for` loops, `forEach` etc.

There are also a few quirks in the script, typically `do` and `delete` do not require space when chaining (I just quickly skimmed through the changes, so I may have missed more).

mbroadst:
@neumino again this is all a matter of taste, but the rules used here are just a run of "fixjsstyle" from the closure-linter project. So these are the rules run for e.g. every single google javascript in existence today (to address your point about quirks). I think the important part is that style is consistent one way or another, which was not the case previously in the project.

I'm sure we can play around with the linter to get something close to your particular style, but the important part of this is to encode the rules and make sure they are run consistently (this is why the jshint target was added, and is forced to run before coverage and test targets)

mbroadst:
@neumino also, are you saying above that you're agains forEach? Why? There were a number of cases in your code that I identified cases where loop variables were used and not hoisted, potentially leading to silently overwritten values. I find the forEach syntax far more expressive, and in most cases does away with the need for a needless loop variable. Are you concerned that it reduces performance or something? Or just have an affinity for for-loops?

neumino:
Ok I thought a bit about that (lint, cleaning and sutff)

I have to run lint at work, and it's kind of a pain. It make sense there because we are many to work on the same code. So far I've been maintaining thinky mostly by myself with the help of a few like @marshall007 or @mindjuice. 
I still update it, but do it for fun (I'm not paid here) and running lint (or something similar) would remove some of the fun.

If you want to submit a PR with adding missing colon I will merge it. If it's about changing how spacing is done, or replacing `for` loops with forEach, I won't merge it.

Sorry for that, I'm aware that you put some effort for that, and I'm genuinely thankful, but as long as I'm the main contributor, I don't want to have to change my style. If one day someone become a recurrent contributor, I would be open to talk about style again, but not yet.

mbroadst:
@neumino I've decided to close my PRs because I think we just differ on opinion a little bit, and that's okay! I understand you're doing this for fun, and genuinely appreciate the work that you've done on this project. For me, I'm contemplating using this as a core part of product at a pretty large scale, so it's very important for me that it be as maintainable as possible, and a huge part of that imho is automation of these types of processes.

As you have a very permissive license, I think it's perfectly fine for me to just maintain my own fork of thinky and progress from there. Obviously, if you end up liking anything I implement you are more than willing to incorporate it back into your fork.

Cheers!

marshall007:
@neumino I definitely understand your concerns in regards to linting being intrusive to your development process. It can be quite a pain especially if the lint configuration doesn't reflect they way you're used to writing code.

That being said, when submitting PRs to open-source projects it's quite nice to be able to validate that your code matches the maintainer's existing conventions in an automated way. The `for` vs `forEach` thing is a perfect example of the kind of things it's nice to be able to detect (regardless of your opinion one way or the other).

So, @neumino, if we could configure a linter to match your preferences I think this would be a good idea. I don't think it's all that important right now given that there aren't that many contributors and you still do ~99% of the work, but I would definitely consider it in the long-term and across all of your projects.

PR #233, opened 2015-05-11
refactor(feed-test): clean up feed test

mbroadst:
(depends on #232 , hence the massive change set here)

 - opt for asychrone where possible
 - created a fixture class "test" which encompasses creation and
   subsequent dropping of tables/models, leaving no footprint
   behind on the server tested against
 - renamed file to feed.test.js, changed Makefile to only test files
   with .test.js suffixes by default


neumino:
This looks good to me.
I'm not sure that chai is needed though.

Thanks for this pull request @mbroadst -- and the two others :-)

I'll find some time this week to merge things, write docs, and release that.

mbroadst:
Okay, I'll rebase this to stand on its own. Chai is definitely a matter of opinion, I personally find it more expressive and I think there are issues with assert in some cases. You'll notice I left all the assert.deepEquals statements alone - that's because assert.deepEqual doesn't actually do a proper deepEqual compare (when you deepEqual(doc, {}) that should fail, since {} isn't an instance of Document etc).

neumino:
Same thing as in the other thread.

One thing too is that `deepEqual(doc, {})` should not fail as it doesn't look at the prototype chain. There was a pull request on io.js to change that but it was eventually not merged because it was a major breaking change. On a side note, `util.deepEqual(doc1, doc2)` would always fail with `thinky` if deepEqual was looking at the prototype chain.

marshall007:
@mbroadst are you still planning on pulling out your `TestFixture` stuff in another PR? I like the way you did that and I think [rethinkdbdash](https://github.com/neumino/rethinkdbdash) could benefit from something similar.

mbroadst:
@neumino yes, that's why deepEqual is fundamentally broken, it should be called "superficialCompare" or something because it's not in fact deeply comparing the two objects (ignoring the prototypes)

@marshall007 by all means you can use my code, but at this point I've merged it into my master and will be continuing development over there. There's definitely a strong need for something like that (or something refactored from that), as currently the tests don't really clean up after themselves at all.

PR #236, opened 2015-05-19
Emit document on saved event

### LINKED ISSUE #255: Read query result in batches
I found this in RethinkDB docs:

>If the table contains a large number of documents a query will not return all of them at once, which would saturate the network and/or require a lot of memory on the client. Instead the query will return the results in batches and fetch more data as needed.

I need to set `maxBatchRows` in this case because the default value is set to `unlimited`. Does this work with Thinky?

JSteunou:
should fix #255


JSteunou:
oopsy, wrong issue number

PR #237, opened 2015-05-19
Emit document on saved event

### LINKED ISSUE #235: no id on document in the "saved" callback
```
Post.docAddListener('saved', function(post) {
    console.log('a post has been saved ' + entry.id);
}

// a post has been save undefined
```

1. the doc say "save" but it's "saved"
2. it returns `r.table('post')` and it does not have the attribute I was waiting for

JSteunou:
should fix #235


neumino:
Thanks

PR #263, opened 2015-06-13
fix `this` ref in docs for models pre save hook

chovy:
The docs had a missing reference to `this` in the pre-save hook example:


neumino:
Thanks

Issue #234, opened 2015-05-18
Export schema


neumino:
We can just implement a method `toJSON` on each type.

freak4pc:
+1

kheide:
How is it going with this? I looked at implementing the toPublic define function you mentioned in another issue, but when I am returning an array of results I don't want to iterate over it, add this function and then return it.

neumino:
@kheide -- I'm not sure to understand your question.
What don't you want to do?

nooks:
+1

Tlvenn:
+1

Would be interesting to facilitate adoption of GraphQL with Thinky.

neumino:
Exporting a schema is technically not too hard. What's harder is to reimport it.
Then while I opened this issue, I'm a bit fuzzy on why people would need that and what they would do with that:
- What's the expected output for the schema? How to report properties like min, max etc.
- Should it be JSON?
- Should it include relations?

Tlvenn:
One use case is to interface it with GraphQL for example. GraphQL expects your to build a schema for your app and if you use Thinky, that schema gonna be pretty close to the one you already have declared in Thinky.

Using a fluent GraphQL builder like https://github.com/devknoll/graphql-schema, one could leverage the schema from Thinky to build most of the schema needed for GraphQL.

Graffiti is for example a project which aims to do that and they have an adaptor for Mongo: https://github.com/RisingStack/graffiti-mongoose

It would also be pretty cool to be able to attach some meta on attributes and relations that such tool could leverage. By that I mean, let users attach any meta whatsoever.

As far as GraphQL is concerned, for each attributes, it needs its type, if there is a not null constraint and if there is a default value. I would imagine that for reflection purpose that might be enough also. Not sure either how to best export complex validation rules.

99% of the time, you would want to consume directly such schema in JS, so JS objects would make more sense.

And yes, relations are a must.

There is absolutely no need imho to support the reverse process of re importing a schema export which definitely would be much harder.

dcomcto:
+1 for anything that helps support gaphql

neumino:
Can someone actually write down what exporting a schema means? I can add a type field and then just call `JSON.stringify` on it, but I'm pretty sure this is not what people want.

Also I'm not sure to understand what the point of GraphQL. It's a query language, but I don't see how this help to query RethinkDB. There's an issue to support it, but this is not settle yet. Or maybe I'm missing something about GraphQL?

grantcarthew:
[...]

PR #266, opened 2015-06-16
Exportable schema

davewasmer:
This is an initial pass at exportable schema (see #234). I fully expect that this will not cover all edge cases, but I just wanted to get something up to start the discussion / iterate from. @neumino - does this seem like a good approach? Anything you'd like to see changed?

I'd personally also like to see the ability to export relationships (I'm fine if that's independent of exporting schema), but I'll address that in a separate PR most likely.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -254,6 +254,21 @@ Model.prototype.getTableName = function() {
   return this._getModel()._name;
 }
 
+/*
+ * Return the schema for this Model
+ */
+Model.prototype.getSchema = function() {
+  return this._schema;
+}
+
+/*
+ * Return the exported representation of the schema for this model
+ */
+Model.prototype.exportSchema = function() {
+  return this.getSchema().export();
+}
+
+
 
 Model.prototype.fetchIndexes = function() {
   var self = this;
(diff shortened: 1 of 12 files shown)


davewasmer:
Apologies for the messy diff - I thought I skipped those hunks when adding, but I guess not.

On that thought - @neumino, would you be open to a PR that cleans up the code a bit? I.e. trims whitespace, adds missing semicolons, etc?

neumino:
Hello @davewasmer 

I actually don't remember what the use case for exporting schema was. Why do you need to be able to export a schema?

Your PR currently seems to export a human readable schema, but doesn't export more properties like the maximum length, if the field is required or not etc.
So you probably need `export` basically the whole object in each `type/<type>,js`

davewasmer:
My use case is primarily around reflection. I'm trying to serialize a Thinky model into different JSON formats, without hardcoding per-model serializers. So having the ability to export the schema provides me a way to inspect the model's schema at runtime and dynamically serialize it. If there is a better approach here that I'm missing, let me know :smile: 

Right now, yes, the export is human-readable. Like I mentioned, this is just a proof of concept, it should be easy to modify the `.export()` functions to export whatever we'd like. I just wanted to get your take on the concept first.

neumino:
The basic implementation seems reasonable to me.
Instead of just exporting the type, we should export the whole Type object (with all its attributes + its type).

There's also a need to import the schema after that :)

Tlvenn:
One more use case would be to assist building GraphQL schema.

Tlvenn:
Any chance to see that PR merged ?

neumino:
This PR is kind of incomplete. I'll comment more on the issue.

contra:
+1 - this is also super important for us

neumino:
@contra -- what is it for?

contra:
@neumino Form generation on the front-end and meta endpoints

contra:
How does this deal with relationships? It would be super helpful if those were also part of the export

PR #268, opened 2015-06-23
Summer cleaning

rasapetter:
Here's an attempt to improve code hygiene a bit. I personally prefer small, readable commits, therefore you'll find a few commits that will only change the indentation, cleaning up after their parents, who would've been difficult to read had the indentation change been in the same commit.

This branch is possibly a bit opinionated. Feel free to cherry-pick or ignore entirely.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -263,7 +263,7 @@ Model.prototype.fetchIndexes = function() {
       var r = self._thinky.r;
       var query = new Query(self);
       r.table(self.getTableName()).indexList().do(function(indexes) {
-        return r.table(self.getTableName()).indexWait(r.args(indexes))  
+        return r.table(self.getTableName()).indexWait(r.args(indexes))
       }).run().then(function(indexes) {
         for(var i=0; i<indexes.length; i++) {
           self._getModel()._indexes[indexes[i].index] = true;
@@ -322,7 +322,7 @@ Model.prototype.ensureIndex = function(name, fn, opts) {
           })
         )
       ).run().then(function(result) {
-        self._getModel()._indexes[name] = true; 
+        self._getModel()._indexes[name] = true;
         self._indexWasCreated('local');
 
         resolve();
@@ -516,7 +516,7 @@ Model.prototype.belongsTo = function(joinedModel, fieldDoc, leftKey, rightKey, o
         })
       )
     )
-    
+
     if (self._getModel()._tableCreated === true) {
       query.run().then(function(result) {
         self._indexWasCreated('local');
@@ -585,7 +585,7 @@ Model.prototype.hasMany = function(joinedModel, fieldDoc, leftKey, rightKey, opt
     type: 'hasMany'
   };
   joinedModel._getModel()._localKeys[rightKey] = true;
-  
+
   options = options || {};
   if (options.init !== false) {
     var tableName = joinedModel.getTableName();
@@ -1120,7 +1120,7 @@ Model.prototype._parse = function(data) {
         }
       }, function(error) {
         var newError = new Error("The results could not be converted to instances of `"+self.getTableName()+"`\nDetailed error: "+error.message);
-      
+
         return reject(newError);
       });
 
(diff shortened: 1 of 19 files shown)


marshall007:
:+1: I think the switch statements are much more readable.

neumino:
I was going to say that the switch statements are harder to read :)

I just quickly looked at it, and overall it looks good to me (there's also a fix hidden). I'll take a second look tonight and probably merge it as it is.

rasapetter:
@neumino Hidden fix? Interesting. Where?

rasapetter:
I took the liberty of expanding on this a bit, hope it's ok. I saw an opportunity to get rid of quite a bit of duplicated code in Model.js and it's raining outside...

lib/model.js
```diff
@@ -233,6 +233,21 @@ Model.prototype._tableWasCreated = function() {
...
 /*
+ * Return a promise which resolves when/if the Model is ready to use
+ */
+Model.prototype._hasTable = function() {
+  var self = this;
+  if (this._getModel()._tableCreated) {
+    return new Promise.resolve();
+  }
+  return new Promise(function(resolve) {
+    self.on('created', function() {
```
marshall007: This should probably be `self.once('created', resolve)` so that the listener gets destroyed afterwards.

neumino:
Thanks @rasapetter I think I went through all the commits.
The indentation diff was a kind of annoying to follow, but I did my best and didn't see any issue.

Any chance you can revert the last 3 commits? These implements something that we use to have but was not working when people had too many models/relations.

Thanks for the clean up! (and the fix :-))

rasapetter:
@neumino @marshall007 Agreed. `once` would be better here.

Correct me if I'm wrong @neumino, but doesn't the "Too many listeners" error usually mean that EventEmitter's listener limit is set too low? I.e. that issue could've been resolved by raising the limit using [setMaxListeners()](https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n)?

In any case, I'd like to give this one another shot if you don't mind? I think there's some real readability gains to be made by reducing the amount of duplicated code. I can do it in another branch/PR if you would prefer that.

marshall007:
I think @neumino is talking about the case where you create your model and then immediately try to save *n* documents before the table/indices are created. This will attach *n* `created` listeners to `Model`, so there's not really a reasonable value you could use for `setMaxListeners()` that would work in all cases.

rasapetter:
@marshall007 Well, you could "set to zero for unlimited". That would be similar to what Thinky has today with its `_onTableCreated`: an array of unlimited size containing functions which will be called at a certain point.

marshall007:
@rasapetter oh, good point I didn't realize you could set it to unlimited.

Another option would be to refactor `Thinky.prototype.createModel` slightly to pass in a table creation promise to `Model.new()`. Then `model._hasTable` could just keep returning that same promise.

rasapetter:
@marshall007 That could work. Or perhaps even move the table creation logic into Model.new?

rasapetter:
I'm was thinking something like this: https://github.com/neumino/thinky/compare/master...rasapetter:promisify-table-creation

marshall007:
@rasapetter since you're already catching the db exists error, I think `dbReady` could be simplified a little (not tested):

```js
Thinky.prototype.dbReady = function() {
  var self = this;
  if (this._dbReadyPromise) return this._dbReadyPromise;
  
  var r = self.r;
  return this._dbReadyPromise = r.dbCreate(self._config.db)
  .run().error(function(error) {
    if (error.message.match(/^Database `.*` already exists in/)) return;
    throw error;
  });
};
```

The same is true for `tableReady`. Other than that, this looks solid.

neumino:
@rasapetter -- for now if you can remove the last 3 commits it would be cool. If it's too much work I can cherry pick.

I'm not a big fan of settings `maxListeners` to `Infinity`. While we can't predict how many listeners we need, we shouldn't remove this safety since people can genuinely leak listeners (they do have access to all the EventEmitter methods. If you want to simplify the code, it's probably a better idea to have `tableReady` return a promise like you did in a branch. The code was written at a time when my understanding of promises was pretty poor ^_^

Thanks!

rasapetter:
@neumino No problem at all. I reset the branch to its original state.

neumino:
Thanks @rasapetter, merging now :)

PR #276, opened 2015-06-29
`Model.save(object)` should resolve with single object

### LINKED ISSUE #275: `Model.save(object)` resolves with an array
This might be a regression. I seem to recall seeing a similar issue in the past, but I can't find it.

marshall007:
Fixes #275. After looking at the tests it appears that we were expecting what I would consider the "incorrect" behavior. If the current behavior is intentional feel free to close this and the related issue, just note that it's not [currently documented](http://thinky.io/documentation/api/model/#save) one way or the other right now.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -971,8 +971,9 @@ Model.prototype.execute = function(options) {
 Model.prototype.save = function(docs, options) {
   var self = this;
   var r = self._getModel()._thinky.r;
+  var isArray = Array.isArray(docs);
 
-  if (Array.isArray(docs) === false) {
+  if (!isArray) {
     docs = [docs];
   }
 
@@ -1064,6 +1065,11 @@ Model.prototype.save = function(docs, options) {
       });
     }
   })
+
+  if (!isArray) {
+    return p.get(0);
+  }
+
   return p;
 }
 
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1064,6 +1065,11 @@ Model.prototype.save = function(docs, options) {
       });
     }
   })
+
+  if (!isArray) {
+    return p.get(0);
```
neumino: I just wanted to say that this is pretty neat. I like it.

neumino:
The main reason behind `Model.save` was to allow batch inserts. I don't remember anyone explicitly asking for an array instead of an object in case of a single insert.

Your pull request seems to be the appropriate behavior IMO. Since it's a non-backward compatible change, I'm leaving people a few days before merging. If no objections, I'll merge it as it is since the code is good :)

Thanks again @marshall007 :)

marshall007:
I like using `Model.save` because `document.save` doesn't currently support the `conflict` optarg. By setting `conflict:update` you can avoid the extra step of having to use `document.setSaved()`.

lib/model.js
marshall007: Yea, bluebird is awesome :)

neumino:
Merging since no one has complained :)
Thanks @marshall007

PR #278, opened 2015-07-01
Model promises

rasapetter:
As discussed in #268, moving the table creation code into model.js and making more use of promises would be good for readability and worth the effort. I took a stab at that and went a bit further by eliminating the custom event listener implementations (`onTableCreated` and `onTableReady`) from the file and replaced them with promise based solutions.

I believe model.js should be easier to reason about after these changes. There are more things that could be done (e.g. the `save` and `_parse` functions could use a bit of polishing) but perhaps its best to not do too huge PRs?

A summary of the proposed changes:

* Introduce a `ready()` function which returns a promise that resolves when the Model's table and relations have been successfully created.
* Replace all use of `onTableReady()` and reads of `_tableReady` with `ready()`, both internally and externally.
* Break out index creation to its own function and use it in all relation functions.
* Cleanup.

Some commit messages are admittedly a bit rough around the edges. I'll be happy to clarify any if you find them too brief. Please have a look and tell me if anything looks off.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -758,7 +758,7 @@ Document.prototype._saveSelf = function(
     }
   }
 
-  var saveSelfHelper = function() {
+  self.getModel().ready().then(function() {
     util.tryCatch(function() {
       // Validate the document before saving it
       var promise = self.validate();
@@ -777,9 +777,7 @@ Document.prototype._saveSelf = function(
         }).error(reject)
       }
     }, reject);
-  }
-
-  self.getModel().ready().then(saveSelfHelper);
+  });
 }
 
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -580,17 +580,12 @@ Document.prototype._batchSave = function(executeInsert) {
  * @return {Promise}
  */
 Document.prototype._batchSaveSelf = function(executeInsert) {
-  var model = this._getModel();
+  var self = this;
 
   return new Promise(function(resolve, reject) {
-    if (model._tableReady === true) {
+    self.constructor.ready().then(function() {
       executeInsert(resolve, reject)
-    }
-    else {
-      model._onTableReady.push(function() {
-        executeInsert(resolve, reject)
-      });
-    }
+    });
   })
 }
 
@@ -652,18 +647,11 @@ Document.prototype._saveHook = function(docToSave, saveAll, savedModel, callback
     });
 
     //TODO Remove once
-    if (model._tableReady === true) {
+    self.constructor.ready().then(function() {
       Promise.all(promises).then(function() {
         self._onSavedBelongsTo(copy, docToSave, belongsToKeysSaved, saveAll, savedModel, resolve, reject);
       }).error(reject);
-    }
-    else {
-      model._onTableReady.push(function() {
-        Promise.all(promises).then(function() {
-          self._onSavedBelongsTo(copy, docToSave, belongsToKeysSaved, saveAll, savedModel, resolve, reject);
-        }).error(reject);
-      });
-    }
+    });
   });
   return p.nodeify(callback);
 }
@@ -791,14 +779,7 @@ Document.prototype._saveSelf = function(
     }, reject);
   }
 
-  if (self.__proto__._model._tableReady === true) {
-    saveSelfHelper();
-  }
-  else {
-    model.once('ready', function() {
-      saveSelfHelper();
-    })
-  }
+  self.constructor.ready().then(saveSelfHelper);
 }
 
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -583,7 +583,7 @@ Document.prototype._batchSaveSelf = function(executeInsert) {
   var self = this;
 
   return new Promise(function(resolve, reject) {
-    self.constructor.ready().then(function() {
+    self.getModel().ready().then(function() {
       executeInsert(resolve, reject)
     });
   })
@@ -647,7 +647,7 @@ Document.prototype._saveHook = function(docToSave, saveAll, savedModel, callback
     });
 
     //TODO Remove once
-    self.constructor.ready().then(function() {
+    self.getModel().ready().then(function() {
       Promise.all(promises).then(function() {
         self._onSavedBelongsTo(copy, docToSave, belongsToKeysSaved, saveAll, savedModel, resolve, reject);
       }).error(reject);
@@ -779,7 +779,7 @@ Document.prototype._saveSelf = function(
     }, reject);
   }
 
-  self.constructor.ready().then(saveSelfHelper);
+  self.getModel().ready().then(saveSelfHelper);
 }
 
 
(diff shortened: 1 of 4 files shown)


lib/thinky.js
```diff
@@ -46,43 +48,47 @@ function Thinky(config) {
...
+    r.dbList().contains(self._config.db).do(function(result) {
+      return r.branch(
+        result,
+        {created: 0},
+        r.dbCreate(self._config.db)
+      );
+    }).run().then(resolve).error(function(error) {
+      // The `do` is not atomic, we a concurrent query could create the database
+      // between the time `dbList` is ran and `dbCreate` is.
+      if (error.message.match(/^Database `.*` already exists in/)) {
```
marshall007: As mentioned in the other thread, I think we can avoid the extra `r.dbList()` operation here since we catch the "already exists" error anyway.

There's also no need for the `Promise` constructor. We can just chain everything off of the promise returned from `r.dbCreate().run()`.

lib/model.js
```diff
@@ -176,62 +182,99 @@ Model.new = function(name, schema, options, thinky) {
...
+  var self = this;
+  var model = this._getModel();
+  if (!this._initModel) return Promise.resolve();
+  if (this._tableReadyPromise) return this._tableReadyPromise;
+
+  // Create the table, or push the table name in the queue.
+  var r = model._thinky.r;
+  this._tableReadyPromise = model._thinky.dbReady()
+  .then(function() {
+    return r.tableList().contains(model._name).do(function(result) {
```
marshall007: See my [other comment](https://github.com/neumino/thinky/pull/278/files#r33728944). Same thing, I think we should just assume the table needs to be created and catch the error if it already exists.

lib/document.js
```diff
@@ -580,17 +580,12 @@ Document.prototype._batchSave = function(executeInsert) {
  * @return {Promise}
  */
 Document.prototype._batchSaveSelf = function(executeInsert) {
-  var model = this._getModel();
+  var self = this;
 
   return new Promise(function(resolve, reject) {
-    if (model._tableReady === true) {
+    self.constructor.ready().then(function() {
```
neumino: Thinky has a really special prototype chain. And by special, I mean unique.
`self.constructor` is `self.getModel()`, not `self._getModel()`

neumino: See https://github.com/neumino/thinky/blob/master/lib/document.js#L113

neumino: We swap protos here: https://github.com/neumino/thinky/blob/master/lib/model.js#L106

neumino: One more comment: To be clear this work, but the prototype chain is tricky, and it's better for future generation to call `ready` exactly on the appropriate object so hopefully people will be less confuse if they try to understand things work :)

lib/document.js
```diff
@@ -652,18 +647,11 @@ Document.prototype._saveHook = function(docToSave, saveAll, savedModel, callback
     });
 
     //TODO Remove once
-    if (model._tableReady === true) {
+    self.constructor.ready().then(function() {
```
neumino: Same comment about `constructor` and `_getModel`

lib/document.js
```diff
@@ -791,14 +779,7 @@ Document.prototype._saveSelf = function(
...
 
-  if (self.__proto__._model._tableReady === true) {
-    saveSelfHelper();
-  }
-  else {
-    model.once('ready', function() {
-      saveSelfHelper();
-    })
-  }
+  self.constructor.ready().then(saveSelfHelper);
```
neumino: `saveSelfHelper` was created to avoid redundant code. Now that there's `ready`, there's no need to create a function in the middle of a method.
If you don't have time no worries, it's just more cleaning and we can fix that later in a later PR.

neumino: Same thing about `constructor` and `_getModel` too

lib/model.js
```diff
@@ -176,62 +182,99 @@ Model.new = function(name, schema, options, thinky) {
   return model
 }
 
-Model.prototype.isReady = function() {
-  return this._getModel()._tableReady;
-}
+/**
+ * Create the model's table.
+ * @return {Promise=} Returns a promise which will resolve when the table is ready.
```
neumino: Just wanted to say thank you for also writing these comments. I really appreciate your help :+1:

lib/model.js
```diff
@@ -176,62 +182,99 @@ Model.new = function(name, schema, options, thinky) {
...
+    return Promise.all(self._pendingPromises)
+    .then(function() {
+      var i, allFullfilled = true;
+      for (i=0; i<self._pendingPromises.length; i++) {
+         if (!self._pendingPromises[i].isFulfilled()) {
+          allFullfilled = false;
+          break;
+         }
+      }
+      return allFullfilled ? Promise.resolve() : verifyAll();
```
neumino: That's a pretty interesting pattern. It's neat.

rasapetter:
@marshall007 Right, forgot about that. Does dd460a8 look good?

@neumino Is dd460a8 better?

> That's a pretty interesting pattern. It's neat.

Courtesy of @tomsun :)

neumino:
Ok, I think I went through the whole pull request, besides using `constructor` instead of `_getModel()`, I think everything looks good.

It was quite some work to review all of it, so it's probably even more work to do it, so huge thanks @rasapetter -- The code is cleaner/less redundant/easier to read :-)

marshall007:
Looks good, nice work! :+1:

lib/model.js
marshall007: After taking a closer look, I'm confused as to what this is doing. The `for` loop seems redundant since `Promise.all` will only resolve once all the pending promises are fulfilled. @rasapetter did you mean to use `Promise.settle` instead or am I missing something?

marshall007: Oh, additional promises could have been pushed onto the array after `Promise.all` is called. Tricky, but very clever indeed.

rasapetter: Hmm, we should probably reset the `_pendingPromises` array at some point.

rasapetter:
@neumino Oops, I mistakenly referenced the wrong commit in my last message. Here, 10b6a14, is this better?

MarkHerhold:
Can't wait for this! Awesome work @rasapetter!

MarkHerhold:
@rasapetter you are awesome! `.ready()` just solved a lot of my problems.
@neumino you are awesome too. Thanks for merging this in.

neumino:
Why do you need `ready()` for? You can run a query before the table is ready. Thinky will queue your query.

MarkHerhold:
@neumino I use it before I initially save or query for data.

My use-case involves creating about 8 models, ensuring about double that
many indexes then performing the actual query. With the pr, I yield to
ready() for each model (actually Promise.all()) to ensure they are all
ready before the query.

Before the PR I would often get errors saying the server or table (can't
recall exactly which) wasn't ready and occasionally that the index already
existed. Please note that I wasn't using a callback or yielding to
ensureIndex() before the PR and am still not.

If you really want I can work on a code sample reproducing the issue.

PR #284, opened 2015-07-09
Support unix timestamp dates

JohnyDays:


neumino:
Can you also add a test for saving and a test for parsing to make sure that it works as intended?

JohnyDays:
Done too! :rocket:

PR #285, opened 2015-07-09
Support numeric strings for number type

JohnyDays:


lib/document.js
```diff
@@ -455,6 +455,14 @@ Document.prototype.__makeSavableCopy = function(doc, schema, options, model, r)
       return doc;
     }
   }
+  else if (type.isNumber(schema) && (typeof doc === 'string')) {
+    var numericString = parseInt(doc, 10);
```
neumino: s/parseInt/parseFloat

lib/type/number.js
```diff
@@ -109,6 +109,13 @@ TypeNumber.prototype.validate = function(number, prefix, options) {
     throw new Errors.ValidationError("Validator for the field "+prefix+" returned `false`.");
   }
 
+  if(typeof number === 'string'){
+    var numericString = parseInt(number, 10);
```
neumino: Same thing
s/parseInt/parseFloat

neumino:
Same thing here, it should be parseFloat and not parseInt

lib/type/number.js
```diff
@@ -109,6 +109,13 @@ TypeNumber.prototype.validate = function(number, prefix, options) {
     throw new Errors.ValidationError("Validator for the field "+prefix+" returned `false`.");
   }
 
+  if(typeof number === 'string'){
```
neumino: Can you merge the `if` with the rest of the `if`/`else if` statement

neumino:
Can you also add a test for saving and a test for parsing to make sure that it works as intended?

JohnyDays:
Done, thanks!
Amazing test suite btw

JohnyDays:
Giving it another look, I didn't merge the ifs because we need the converted number to go through the same checks after it's conversion. Am I missing something or?

PR #329, opened 2015-08-24
test(Model): reworked model.js test

mbroadst:
* cleaned up test code to be as asynchronous as possible
 * used chai/chai-as-expected to make tests more readable
 * introduced test-fixture to properly track/clean/drop tables
   throughout all test runs
 * made sure to wait for all table creation before operating on
   those tables

You'll notice these tests take far longer to run. That's because previously thinky's tests did not actually wait for table creation (and frankly did not do any cleanup after test runs). This issue should track the progress of table creation speed (the bottleneck): https://github.com/rethinkdb/rethinkdb/issues/4746

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -1,0 +1,893 @@
+'use strict';
+
+var Promise = require('bluebird'),
+    TestFixture = require('./test-fixture'),
+    expect = require('chai').expect,
+    errors = require('../lib/errors'),
+    util = require('./util');
+
+var test = new TestFixture(),
+    thinky, type;
+
+before(function() {
+  return test.initializeDatabase()
+    .then(function() {
+      thinky = test.thinky;
+      type = thinky.type;
+    });
+});
+after(function() { return test.dropDatabase(); });
+
+describe('createModel', function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should create a new model', function() {
+    var model = thinky.createModel(util.s8(), {
+      id: String, name: String
+    });
+
+    return model.ready()
+      .then(function() { expect(model).to.exist; });
+  });
+
+  it('should check if the table was created', function() {
+    var modelName = util.s8();
+    var model = thinky.createModel(modelName, {
+      id: String, name: String
+    });
+
+    return model.ready()
+      .then(function() {
+        return test.r.tableList().run();
+      })
+      .then(function(result) {
+        expect(result).to.contain(modelName);
+      });
+  });
+
+  it('should create multiple models', function() {
+    var model1 = thinky.createModel(util.s8(), {
+      id: String, name: String
+    });
+
+    var model2 = thinky.createModel(util.s8(), {
+      id: String, name: String
+    });
+
+    return Promise.all([ model1.ready(), model2.ready() ])
+      .then(function() {
+        expect(model1).to.not.eql(model2);
+      });
+  });
+
+  it('should throw error for non existent table access', function() {
+    var model = thinky.createModel("nonExistentTable", {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    return expect(model.get(1).run())
+      .to.eventually.be.rejectedWith(/^Table `.*` does not exist in/);
+  });
+});
+
+describe('[_]getModel', function() {
+  it('_getModel', function() {
+    var name = util.s8();
+    var model = thinky.createModel(util.s8(), {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    expect(model._getModel()).to.have.property('_name');
+  });
+
+  it('getTableName', function() {
+    var modelName = util.s8();
+    var model = thinky.createModel(modelName, {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    expect(model).to.have.property('getTableName');
+    expect(model.getTableName()).to.equal(modelName);
+  });
+});
+
+describe('Model', function() {
+  before(function() {
+    test.models.Model = thinky.createModel(util.s8(), { str: String });
+    return test.models.Model.ready();
+  });
+  after(function() { return test.dropTables(); });
+  afterEach(function() { return test.clearTables(); });
+
+  it('should create a new instance of a model', function() {
+    var str = util.s8();
+    var doc = new test.models.Model({ str: str });
+
+    expect(util.isPlainObject(doc)).to.be.true;
+    expect(doc.str).to.eql(str);
+  });
+
+  it('should create multiple instances from the same document', function() {
+    var str = util.s8();
+    var num = util.random();
+
+    var values = { str: str, num: num };
+    var doc = new test.models.Model(values);
+    var otherDoc = new test.models.Model(values);
+
+    expect(doc).to.eql(values);
+    expect(doc).to.not.eql(otherDoc);
+
+    doc.str = doc.str + util.s8();
+    expect(doc.str).to.not.eql(otherDoc.str);
+
+    var anotherDoc = new test.models.Model(values);
+    expect(anotherDoc).to.not.eql(otherDoc);
+    expect(anotherDoc).to.not.eql(doc);
+  });
+
+  it('should create two distinct instances with the same options', function() {
+    var options = { str: util.s8() };
+    var doc1 = new test.models.Model(options);
+    var doc2 = new test.models.Model(options);
+
+    expect(doc1).to.not.eql(doc2);
+  });
+
+  it('should ensure two instances are different', function() {
+    var str1 = util.s8(), str2 = util.s8();
+    var doc1 = new test.models.Model({ str: str1 });
+    expect(doc1.str).to.equal(str1);
+
+    var doc2 = new test.models.Model({ str: str2 });
+    expect(doc2.str).to.equal(str2);
+    expect(doc1.str).to.equal(str1);
+    expect(doc1).to.not.eql(doc2);
+  });
+
+  it('two instances should have different prototypes', function() {
+    var doc1 = new test.models.Model({ str: util.s8() });
+    var doc2 = new test.models.Model({ str: util.s8() });
+
+    expect(Object.getPrototypeOf(doc1)).to.not.eql(Object.getPrototypeOf(doc2));
+  });
+
+  it('two instances should have the same model', function() {
+    var doc1 = new test.models.Model({ str: util.s8() });
+    var doc2 = new test.models.Model({ str: util.s8() });
+
+    expect(doc1.getModel()).to.eql(doc2.getModel());
+  });
+
+  it('docs from different models should not interfere', function() {
+    var str = util.s8();
+    var doc = new test.models.Model({str: str});
+
+    var otherName = util.s8();
+    test.models.OtherModel = thinky.createModel(otherName, {
+      str: String
+    });
+
+    return test.models.OtherModel.ready()
+      .then(function() {
+        var otherStr = util.s8();
+        var otherDoc = new test.models.OtherModel({ str: otherStr });
+
+        expect(doc.str).to.eql(str);
+        expect(otherDoc.str).to.eql(otherStr);
+
+        expect(otherDoc.getModel()).to.not.eql(doc.getModel());
+
+        // TODO: better way to access schema data, maybe Model.getSchema()
+        expect(doc.getModel().getTableName()).to.eql(test.models.Model._schema._model._name);
+        expect(otherDoc.getModel().getTableName(), otherName);
+      });
+  });
+
+});
+
+describe("Batch insert", function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should insert a single document', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Model.save({ id: 'foo' });
+      })
+      .then(function(result) {
+        // NOTE: deepCopy to remove prototype. Otherwise these are
+        //       not, in fact, equivalent objects, even though
+        //       assert will claim they are
+        expect(util.deepCopy(result)).to.eql({ id: 'foo' });
+      });
+  });
+
+  it('should insert a batch of documents', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    var docs = [];
+    for (var i = 0; i < 10; i++) {
+      docs.push({ num: i });
+    }
+
+    return Model.ready()
+      .then(function() {
+        return Model.save(docs);
+      })
+      .then(function(result) {
+        expect(result).to.eql(docs);
+
+        for (i = 0; i < 10; i++) {
+          expect(docs[i].id).to.be.a('string');
+          expect(docs[i].isSaved()).to.be.true;
+        }
+      });
+  });
+
+  it('should validate fields before saving', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Promise.all([
+          expect(Model.save([{id: 4}]))
+            .to.eventually.be.rejectedWith(errors.ValidationError),
+          expect(Model.save([{id: 4}]))
+            .to.eventually.be.rejectedWith("One of the documents is not valid. Original error:\nValue for [id] must be a string or null.")
+        ]);
+      });
+  });
+
+  it('should properly error if a single insert fails', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    var savePromise =  Model.save([{id: '4'}])
+      .then(function(result) {
+        expect(result[0].id).to.equal('4');
+
+        var docs = [];
+        for (var i = 0; i < 10; i++) {
+          docs.push({ num: i, id: '' + i });
+        }
+
+        return Model.save(docs);
+      });
+
+    return Model.ready()
+      .then(function() {
+        return expect(savePromise)
+          .to.eventually.be.rejectedWith(/An error occurred during the batch insert/);
+      });
+  });
+});
+
+describe('Insert', function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should generate savable copies', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      location: type.point()
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Model.save({ id: "foo", location: [1, 2] });
+      })
+      .then(function(result) {
+        expect(result.id).to.equal('foo');
+        expect(result.location.$reql_type$).to.equal('GEOMETRY');
+      });
+  });
+
+  it('should handle options (update)', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Model.save({ id: 'foo' });
+      })
+      .then(function(result) {
+        expect(result.id).to.equal('foo');
+        return Model.save({ id: 'foo', bar: 'buzz' }, { conflict: 'update' });
+      })
+      .then(function(result) {
+        expect(util.deepCopy(result)).to.eql({ id: 'foo', bar: 'buzz' });
+      });
+  });
+
+  it('should handle options (replace)', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Model.save({ id: 'foo', bar: 'buzz' });
+      })
+      .then(function(result) {
+        expect(result.id).to.equal('foo');
+        return Model.save({ id: 'foo' }, { conflict: 'replace' });
+      })
+      .then(function(result) {
+        expect(util.deepCopy(result)).to.eql({ id: 'foo' });
+      });
+  });
+});
+
+describe("Joins", function() {
+  var types = [ 'hasOne', 'belongsTo', 'hasMany', 'hasAndBelongsToMany'];
+  types.forEach(function(type) {
+    describe(type, function() {
+      afterEach(function() { return test.dropTables(); });
+
+      it('should throw if it uses a field already used by another relation', function() {
+        var Model = thinky.createModel(util.s8(), { id: String }, { init: false });
+        var OtherModel = thinky.createModel(util.s8(), {
+          id: String,
+          otherId: String
+        }, {
+          init: false
+        });
+
+        var AnotherModel = thinky.createModel(util.s8(), {
+          id: String,
+          otherId: String
+        }, {
+          init: false
+        });
+
+        Model[type](OtherModel, "otherDoc", "id", "otherId", { init: false });
+
+        return Promise.all([ Model.ready(), OtherModel.ready(), AnotherModel.ready() ])
+          .then(function() {
+            expect(function() {
+              Model[type](AnotherModel, "otherDoc", "id", "otherId");
+            }).to.throw("The field `otherDoc` is already used by another relation.");
+          });
+      });
+
+
+      it('belongsTo should throw if the first argument is not a model', function() {
+        var Model = thinky.createModel(util.s8(), { id: String }, { init: false });
+
+        return Model.ready()
+          .then(function() {
+            expect(function() {
+              Model[type](function() {}, "otherDoc", "otherId", "id");
+            }).to.throw('First argument of `' + type + '` must be a Model');
+          });
+      });
+
+      it('should create an index on the other model', function() {
+        var Model = thinky.createModel(util.s8(), {
+          id: String,
+          foreignKeyName: String
+        });
+
+        var foreignKey = util.s8();
+        var schema = { id: String };
+        schema[foreignKey] = String;
+        var OtherModel = thinky.createModel(util.s8(), schema);
+
+        Model[type](OtherModel, "otherDoc", "modelId", foreignKey);
+        return Promise.all([ Model.ready(), OtherModel.ready() ])
+          .then(function() {
+            return Promise.all([
+              test.r.table(OtherModel.getTableName()).indexList().run(),
+              test.r.table(OtherModel.getTableName()).indexWait(foreignKey).run()
+            ]);
+          });
+      });
+
+      it('should throw on attempted use of _apply', function() {
+        var Model = thinky.createModel(util.s8(), { id: String, notid1: String }, { init: false });
+        var OtherModel = thinky.createModel(util.s8(), { id: String, notid2: String }, { init: false });
+
+        return Promise.all([ Model.ready(), OtherModel.ready() ])
+          .then(function() {
+            expect(function() {
+              Model[type](OtherModel, "_apply", "notid1", "notid2", { init: false });
+            }).to.throw("The field `_apply` is reserved by thinky. Please use another one.");
+          });
+      });
+
+    });
+  });
+
+  describe('hasOne', function() {
+    afterEach(function() { return test.dropTables(); });
+
+    it('should save the join', function() {
+      var Model = thinky.createModel(util.s8(), { id: String });
+      var OtherModel = thinky.createModel(util.s8(), { id: String, otherId: String });
+      Model.hasOne(OtherModel, "otherDoc", "id", "otherId");
+
+      return Promise.all([ Model.ready(), OtherModel.ready() ])
+        .then(function() {
+          expect(Model._getModel()._joins.otherDoc).to.exist;
+        });
+    });
+
+  });
+
+  describe('hasAndBelongsToMany', function() {
+    afterEach(function() { return test.dropTables(); });
+
+    it('should create an index on this table', function() {
+      var Model = thinky.createModel(util.s8(), { id: String, notid1: String });
+      var OtherModel = thinky.createModel(util.s8(), { id: String, notid2: String });
+
+      Model.hasAndBelongsToMany(OtherModel, "otherDocs", "notid1", "notid2");
+
+      var linkName;
+      if (Model.getTableName() < OtherModel.getTableName()) {
+        linkName = Model.getTableName() + "_" + OtherModel.getTableName();
+      } else {
+        linkName = OtherModel.getTableName() + "_" + Model.getTableName();
+      }
+
+      return Promise.all([ Model.ready(), OtherModel.ready() ])
+        .then(function() {
+          return Promise.all([
+            test.r.table(Model.getTableName()).indexList().run(),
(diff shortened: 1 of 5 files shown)


mbroadst:
I forgot to mention: these tests are _basically_ the same as in model.js, however I've reduced a lot of copy-pasta, cleaned it up a bit, etc. I wanted to keep it in a separate file so we could discuss any potential issues (as there were a few last time).

My goal is to come to some conclusions regarding converting the rest of test suite over to some sort of consistent style/format. I would also absolutely love if we could come to an agreement on a .jshintrc that could be added to the project

marshall007:
> - made sure to wait for all table creation before operating on those tables

I think we should probably still have some tests (maybe all?) that don't wait for `Model.ready()`, especially if that's a lot faster. Since thinky is supposed to support executing queries on a model before the table is ready, it's useful to include that behavior in the tests.

mbroadst:
@marshall007 yeah, I agree, and that's possible with passing in `{ init: false }` generally. The real problem with the current design imho is that far too much happens in thinky as a side effect: there's a distinct lack of separation between declaring what your schema looks like and the actual creation of that schema in the database (an example of this is the implicit promises set into motion as a part of creating an association, an internal table schema is automatically created and executed). 

The tests where I added the `ready()` calls here all at some point depend on the table being ready (especially in order to clean it up) - the speed was coming from a sort of set-and-forget mentality. In fact, I had opened another issue against rethinkdb because tearing down the table after the existing queries causes tables to be orphaned with the current version of rethink.

Finally, considering these are the model tests there is probably just going to be more of this type of behavior, rather than the query tests for instance where we might set up a bunch of sample data and just query away without needing to setup/teardown tables. 


(except for with e.g. hasAndBelongsToMany type associations). TH

marshall007:
@mbroadst agreed concerning the "fire-and-forget" stuff, it's nice to have the tests encapsulated the way you've done with `test/test-fixture.js`. That said, I still don't see the value in waiting on `Model.ready()` everywhere (specifically in the `Batch insert` and `Insert` tests).

mbroadst:
@marshall007 well to be honest, adding the `ready()` there is just a way to inform the reader of the test that that is exactly what's happening. In reality, or at least my experience so far, you will wait just as long since either way rethinkdb has to make the table in order for you to insert into it. I haven't seen any speed improvement (at least in Batch insert) by removing the calls to `ready()`. 

By all means fork this PR and improve upon it, I'm very open to discussion.

neumino:
I don't have much time today (and probably this week), so I just gave the pull request a quick look but I'm not sure to understand what this PR is suppose to bring.

A few quick/random thoughts:
- We used to create one model per test, and the truth is while it was prettier, it blows because it takes forever to run the tests. Typically this branch needs 2m40 instead of one minute to run the tests on wercker.
- `chai` is not needed. It's one more dependence, and more verbose tests...
- The tests do not wait for the table to be ready because thinky does it under the hood.
- Test fixtures is a good idea. `initializeDatatabase` has no reason to return a promise though
- Tests are not strictly the same in my opinion. I'll write more comments on the code later.

I'm aware that tests are messy, that we could refactor some code (like test-fixtures), but this PR doesn't seem to simplify things or make developing easier.

mbroadst:
To answer those points in order:

  * I haven't created any more models than exist in `test/model.js`

  * I have no idea why you are so against chai :) it _is_ an extra dependency, but only for the tests (this will have zero effect on the end user), and it in fact makes the tests far more readable in some cases (no more try/catch + done(), instead I can use chai-as-promised to write a single human-readable string to tell me what the test is doing).

  * thinky is not complete in it's support for waiting for the table, as evidenced by my need to add it everywhere to ensure that tables were properly created before trying to remove them. Also, as I've mentioned above adding the table ready here makes it very explicit (a _good_ thing) to the user that the table is ready before trying to operate on the table. The fact that thinky (through side-effects) does this stuff in the background I think is a _bad_ thing, because its more cognitive overhead for the implementor. Adding `ready()` where it's been added has a beneficial effect in some places, and no effect in other places, but it is always very clear that it's a required step this way. 

   * okay, sure

   * The tests are definitely the same. I removed a number of duplicate tests, and even added a few extra tests for completeness sake.

As for the comment "doesn't seem to simplify things or make developing easier", can you please expand upon that? Because I strongly disagree.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+describe('createModel', function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should create a new model', function() {
+    var model = thinky.createModel(util.s8(), {
+      id: String, name: String
+    });
+
+    return model.ready()
+      .then(function() { expect(model).to.exist; });
```
neumino: `model` is immediately defined. There's no need to wait for the promise returned by `ready`

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+
+  it('should throw error for non existent table access', function() {
+    var model = thinky.createModel("nonExistentTable", {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    return expect(model.get(1).run())
+      .to.eventually.be.rejectedWith(/^Table `.*` does not exist in/);
```
neumino: This is kind of insane to me. Why `rejectedWith` and not `rejected.with`?
Why not `toBeRejectedWith`?

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+describe('[_]getModel', function() {
+  it('_getModel', function() {
+    var name = util.s8();
+    var model = thinky.createModel(util.s8(), {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    expect(model._getModel()).to.have.property('_name');
```
neumino: As far as I can tell from a small test, this is equivalent to

```
assert('_name' in model._getModel()))
```

Which is not the same as

```
assert(model._getModel().hasOwnProperty('name')))
```

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+
+  it('getTableName', function() {
+    var modelName = util.s8();
+    var model = thinky.createModel(modelName, {
+      id: String, name: String
+    }, {
+      init: false
+    });
+
+    expect(model).to.have.property('getTableName');
```
neumino: In the `getTableName`, I actually test a verify specific thing: The location of the `getTableName` property.
The proto chain of `thinky` objects is unusual and tricky, hence the test.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+
+  it('should create multiple instances from the same document', function() {
+    var str = util.s8();
+    var num = util.random();
+
+    var values = { str: str, num: num };
+    var doc = new test.models.Model(values);
+    var otherDoc = new test.models.Model(values);
+
+    expect(doc).to.eql(values);
```
neumino: This is different than testing `doc === values`.
It's actually an unusual equality test from what I could test.

My guess is without looking in the code, it does something like this
```
for(key in doc) {
  if (doc[key] != values[key]) { // they probably recurse here instead of checking !=
    throw ...
  }
}
```

Meaning that
```
var a = {foo: "bar"}
var b = {};
b.__proto__.foo = "bar"
expect(a).to.eql(b) // doesn't throw...
```

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+    expect(doc).to.not.eql(otherDoc);
```
neumino: I don't know why this pass. I guess this is because of the emitters on Document

neumino: At least I'm pretty sure that it doesn't check for strict equality

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+
+    doc.str = doc.str + util.s8();
+    expect(doc.str).to.not.eql(otherDoc.str);
+
+    var anotherDoc = new test.models.Model(values);
+    expect(anotherDoc).to.not.eql(otherDoc);
```
neumino: Same thing for these two lines.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+    expect(anotherDoc).to.not.eql(doc);
+  });
+
+  it('should create two distinct instances with the same options', function() {
+    var options = { str: util.s8() };
+    var doc1 = new test.models.Model(options);
+    var doc2 = new test.models.Model(options);
+
+    expect(doc1).to.not.eql(doc2);
```
neumino: Ditto

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+
+  it('should ensure two instances are different', function() {
+    var str1 = util.s8(), str2 = util.s8();
+    var doc1 = new test.models.Model({ str: str1 });
+    expect(doc1.str).to.equal(str1);
+
+    var doc2 = new test.models.Model({ str: str2 });
+    expect(doc2.str).to.equal(str2);
+    expect(doc1.str).to.equal(str1);
+    expect(doc1).to.not.eql(doc2);
```
neumino: Ditto for all the other `to.eql`/`to.not.eql` on documents

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+  });
+
+  it('two instances should have different prototypes', function() {
+    var doc1 = new test.models.Model({ str: util.s8() });
+    var doc2 = new test.models.Model({ str: util.s8() });
+
+    expect(Object.getPrototypeOf(doc1)).to.not.eql(Object.getPrototypeOf(doc2));
```
neumino: Why `Object.getPrototypeOf(doc)` instead of `doc.__proto__`?
This is node.js/io.js code, there's no need to worry about browser compliances.

test/model.test.js
mbroadst: I think you've made your point with these, they can all be replaced with 'equal' (http://chaijs.com/api/bdd/ `equal` section indicates a strict === equality). Problem solved

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+  it('docs from different models should not interfere', function() {
+    var str = util.s8();
+    var doc = new test.models.Model({str: str});
+
+    var otherName = util.s8();
+    test.models.OtherModel = thinky.createModel(otherName, {
+      str: String
+    });
+
+    return test.models.OtherModel.ready()
```
neumino: thinky doesn't require you to wait to create object. That's why `thinky()` doesn't return a promise but a plain object.

test/model.test.js
mbroadst: Because its frankly kind of weird to use it, this is actually the first time I've seen someone use it as judiciously as you do.

see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto

specifically: "Warning: While Object.prototype.__proto__ is supported today in most browsers, its existence and exact behavior has only been standardized in the ECMAScript 6 specification as a legacy feature to ensure compatibility for web browsers. For better support, it is recommended that only Object.getPrototypeOf() be used instead."

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+describe("Batch insert", function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should insert a single document', function() {
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
```
neumino: Ditto here and in other places. There's no need to wait for the document to be ready. It's miss leading if someone reads the tests to learn how thinky behaves.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      .then(function() {
+        return Model.save({ id: 'foo' });
+      })
+      .then(function(result) {
+        // NOTE: deepCopy to remove prototype. Otherwise these are
+        //       not, in fact, equivalent objects, even though
+        //       assert will claim they are
+        expect(util.deepCopy(result)).to.eql({ id: 'foo' });
```
neumino: Ditto on the equality operator of chai on Object. We expect a strict equality here.

neumino: Apologies, I read the wrong test. But this is still the same issue with `to.eql`. It's strictly different than `assert.deepEqual`.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      });
+  });
+
+  it('should insert a batch of documents', function() {
```
neumino: It's missing `done` here. The test is synchronous. If the test fails, it may print something like "Uncaugh Exception...`, but the test will still pass

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+        expect(result).to.eql(docs);
+
+        for (i = 0; i < 10; i++) {
+          expect(docs[i].id).to.be.a('string');
+          expect(docs[i].isSaved()).to.be.true;
+        }
+      });
+  });
+
+  it('should validate fields before saving', function() {
```
neumino: Ditto...

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+    var Model = thinky.createModel(util.s8(), {
+      id: String,
+      num: Number
+    });
+
+    return Model.ready()
+      .then(function() {
+        return Promise.all([
+          expect(Model.save([{id: 4}]))
```
neumino: You are running twice Model.save here, not sure why.
I'm also not sure why you need to wrap things in Promise.all?

neumino: Though I'm not sure what `expect(...).to.be.eventually.rejectedWith` is supposed to return

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+        return Promise.all([ Model.ready(), OtherModel.ready(), AnotherModel.ready() ])
+          .then(function() {
+            expect(function() {
+              Model[type](AnotherModel, "otherDoc", "id", "otherId");
+            }).to.throw("The field `otherDoc` is already used by another relation.");
+          });
+      });
+
+
+      it('belongsTo should throw if the first argument is not a model', function() {
```
neumino: type+' should throw...'

test/model.test.js
mbroadst: This is exactly what I think is wrong about what you two are saying. This is _exactly_ how thinky behaves, except you seem to want it to be a side-effect of createModel so it just sort of "magically" happens in the background. (reference: https://github.com/neumino/thinky/blob/master/lib/model.js#L161) 

I would much prefer to explicitly write this out so I know exactly what's happening. More concerning cases (brought up during my testing) were that in the event of creating cross tables for association whole models are created in the background without much control offered to the user and then even more promises are set off without any possible way to interact with them.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+          foreignKeyName: String
+        });
+
+        var foreignKey = util.s8();
+        var schema = { id: String };
+        schema[foreignKey] = String;
+        var OtherModel = thinky.createModel(util.s8(), schema);
+
+        Model[type](OtherModel, "otherDoc", "modelId", foreignKey);
+        return Promise.all([ Model.ready(), OtherModel.ready() ])
```
neumino: Just just need `OtherModel` to be ready. If we need to wait for `Model.ready`, there's a bug.

test/model.test.js
mbroadst: I say leave it up to the user. Your documentation doesn't recommend this style, but for careful testing I think the verbosity is very important to ensure that certain operations occur in the correct order.

test/model.test.js
mbroadst: it returns a promise. I'm running it twice to test both that it throws a `ValidationError` as well as that particular string. It makes no difference running save once or twice, although arguably its enough to just test that it throws a `ValidationError`

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+          .then(function() {
+            return Promise.all([
+              test.r.table(OtherModel.getTableName()).indexList().run(),
+              test.r.table(OtherModel.getTableName()).indexWait(foreignKey).run()
+            ]);
+          });
+      });
+
```
neumino: For `hasAndBelongsToMany`, there's a test to check for an index on the Model that called `hasAndBelongsToMany`.

neumino: It's also missing the test for the linked table.

neumino: Never mind, it's below.

test/model.test.js
marshall007: I think this and other similar cases are ok. For convenience, `mocha` allows you to return a promise from your test so you don't have to deal with the `done` callback.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      }).then(function(docs) {
+        expect(docs[0].other.bar).to.be.true;
+      });
+  });
+});
+
+describe('ensureIndex', function() {
+  afterEach(function() { return test.dropTables(); });
+
+  it('should add and ensure an index', function() {
```
neumino: done?

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      })
+      .then(function(result) {
+        return Model.orderBy({ index: "num" }).run();
+      })
+      .then(function(result) {
+        expect(result[0].num).to.equal(42);
+      });
+  });
+
+  it('should add an index with multi', function() {
```
neumino: done?

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      .then(function(result) {
+        expect(result).to.have.length(1);
+        return Model.getAll(2, { index: 'nums' }).run();
+      })
+      .then(function(result) {
+        expect(result).to.have.length(1);
+      });
+  });
+
+  it('should accept index options', function() {
```
neumino: done?

test/model.test.js
mbroadst: no bug, this is just the requirement of actually cleaning up the tables after running the tests (as well as a continued reflection on the side-effecty nature of thinky). If you call `thinky.createModel` without `{ init: false }`, then it will automatically make the model for you as an unspecified time in the future (depending on the micro task queue). The only way to safely guarantee cleanup is to wait for all tables to be created.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+      numVirtual: {
+        _type: 'virtual'
+      }
+    }, { init: false });
+
+    var doc = new Model({ num: 1 });
+    return doc.validate();
+  });
+
+  it('should not save virtual fields', function() {
```
neumino: done?

test/model.test.js
mbroadst: this is correct, see "Working with Promises" @ https://mochajs.org/

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+    return Model.ready()
+      .then(function() {
+        return doc.save();
+      })
+      .then(function(result) {
+        expect(result.numVirtual).to.equal(3);
+        return Model.get(1).execute();
+      })
+      .then(function(result) {
+        expect(result.numVirtual).to.be.undefined;
```
neumino: This is insane to me. While I can see how this work, I believe most of the people don't know how this could `to.be.undefined` could work.
I don't see how this is a usual pattern in JS.

test/model.test.js
```diff
@@ -0,0 +1,893 @@
...
+        return Model.get(1).run();
+      })
+      .then(function(result) {
+        expect(result.numVirtual).to.equal(3);
+      });
+  });
+
+  it('should not save virtual fields, but should revert them (if no default)', function() {
```
neumino: done?

mbroadst: please stop writing `done` until you read more about mocha :smile:

neumino:
I don't have much time tonight, but still did a full pass (I think) on the test file.

Thanks @mbroadst, I appreciate the effort but there are a few issues with the tests. They are not strictly equivalent (or at least testing the same set of things). The prototype chain in thinky is really special, and the tests were actually meaningful.

The comments about equality, `to.be.undefined` etc. are one reason I don't like chai. I don't want to add a library for test that does fancy things like `to.be.undefined` but that no one knows how it works or what it tests.

I'm still the one who has to run the tests the most (since I'm still the main one writing code for thinky), and honestly I don't want to have to google how chai works everytime or write a script to be sure that it will test what I want.

The extra time it needs to run the tests may not be an issue for most of the people, but as the one who always run the tests before releasing and often while adding features, this would be a huge pain/downside for my interest to work on thinky.

neumino:
Arg you guys already replied to my comments. I have to run now. I'll try to catch up tonight (if it's not too late) or tomorrow.

test/model.test.js
marshall007: @mbroadst in production, do you wait for all your models to be ready before starting the server? I've found that the way `thinky` does things significantly reduces application startup time. You can start the server and immediately start accepting requests. If some of those requests hit the database, `thinky` just waits until the relevant table(s) are ready before executing your queries. This is particularly important if you're deploying code changes that create new indices on a large table.

I imagine this is how most people use `thinky`, so to mirror what @neumino said, I think it's important that the tests reflect that.

mbroadst:
afaict the issues you brought up are quite easily solved. your discomfort with chai can be overcome (I'll just revert to assert, even though I think it's stretching it a bit far to say that nobody reading chai would have any idea what it's doing.. "1,006,653 downloads in the last month" on npmjs just _completely_ disagrees with you here). 

As far as the time it takes to run the tests, well that's just how long it has to take until the guys at rethink optimize the tableCreate process I think. IMHO an ORM/ODM has to be incredibly precise in its testing, specifically involving making sure that tests run from a clean slate and in a reproducible manner. I'm adding roughly 1min (in fact I think its more like ~40s) to your test run here, but you make it seem like we're talking about 10min extra here (and tbh if it _was_ 10min I'd still be for it).

marshall007:
> The extra time it needs to run the tests may not be an issue for most of the people ...

@mbroadst if the database/table cleanup is adding additional overhead perhaps, we could add a command line argument to disable this? That way the CI server can run in "no-cleanup" mode for performance and/or @neumino's sanity :)

test/model.test.js
mbroadst: @marshall007 I do. In fact, I think it's frankly a bit irresponsible (in production) to depend on my ORM/ODM to even create tables for me. This is where my suggestion that we split the schema/model definition from the actual table creation into parts came from. If I need to update indices on a huge table, I don't do that as part of a rollout of my node app. Instead, I run those operations explicitly to ensure data integrity (and that the time it takes to run such an operation doesn't affect user experience).  

I don't disagree that running `ready()` here is different than the form suggested in the documentation, but could you please recommend an alternative that allows for cleaning up that table then?

mbroadst:
@marshall007 we certainly could, though I believe it would be non-trivial.

Again I point to Sequelize. They release regularly with only two official maintainers, and a test suite on [travis](https://travis-ci.org/sequelize/sequelize) that runs 25 variations for a grand total of 52 min 23 sec of testing. It is literally bizarre to me that we are discussing how `expect(undefined).to.be.undefined` is "insane", while we are saying that a 2 minute overall test time is far too great to handle for release purposes.

test/model.test.js
marshall007: @mbroadst I see what you're saying, that makes sense. It would probably simplify things, actually if `thinky` could assume the tables were already provisioned before the user calls `.run()`. Then we could throw all the table provisioning logic under `Model.provision()` (or something) and users would be instructed to ensure their tables were provisioned as part of their deployment process.

That said, I still think it's better fit for a separate discussion and the tests should continue to reflect *documented* best practices.

> but could you please recommend an alternative that allows for cleaning up that table then?

In `TestFixture`, you're already looping over `thinky.models`. I think it could just look something like this (untested, but notice the `model.ready()`):

```js
TestFixture.prototype.dropTables = function() {
  var self = this,
      keys = Object.keys(self.thinky.models);

  return Promise.map(keys, function(key) {
    var model = self.thinky.models[key];

    if (model._initModel) {
      return model.ready().then(function () {
        return self.r.tableDrop(key).run();
      });
    }
  }).then(function() { self.thinky._clean(); });
};
```

neumino:
Here's a list of actionable items for this PR:
- Please add `done` in the argument even if it works when you return a promise. Typically on line 742, one could be lead into thinking that `doc.validate()` returns a promise. While it can be the case in some situations, it's not here.
- Please remove `chai`. It's an extra dependencies that is not required. `assert` is part of the core, works fine and I happen to be familiar with it. I'm not familiar with chai, and since I still maintain the code, it's more work for me. An example of why I think chai doesn't do a good job is the description of `eql`. It says "assert that the target is deeply equal to value. It's typically different than `assert.deepEqual` since it tests for all keys in the object and its prototype chain
http://chaijs.com/api/bdd/#eql
- Have the fixture be able to create a model and store whether the table under the hood is ready or not.
Have this method takes an extra argument to match `{init: false}`. Model should be recreated, tables should not. Tests should take the same time or be faster since we currently recreate tables per "group" of tests (`describe`) if I'm not mistaken.
- Fix the tests to test the same things. You can add extra tests, the new test should supersede the old test. The tests about strict equality, deep equality, etc. are important.
- Do not wait for model to be ready to fire queries. Queries are queued until the model is ready, this is by design. It's dev friendly. In prod, it is supposed to be a no op. If it is not, you are probably crazy-ish. In prod you are supposed to pass `{init: false}` and run a dummy script with your models to create the indexes (though I admit I never wrote about that). Typically this use case sucks if you wait for your model to be ready:
- Have a dataset/models already in the database
- Create a new experimental relation between two models (and therefore a new index) that is not required for real traffic, only for experiments.
- Launch the index creation
- Kill/restart your node apps
Now your app won't start serving until your index is ready. In this case you push all your traffic on a few existing nodes (if some didn't die), or you have no node to serve the traffic.


Then it should be good I think. I don't really like having a `for` loop to generate test because it's more painful to isolate the test after when it fails, but that's fine, I can work around.

mbroadst:
> Please add done in the argument even if it works when you return a promise. Typically on line 742, one > could be lead into thinking that doc.validate() returns a promise. While it can be the case in some
> situations, it's not here.

you clearly saw that as needless [here](https://github.com/neumino/thinky/blob/master/test/model.js#L739). Sure in cases like 742 I shouldn't presently return the validation function (though I believe in the future validations should be a promise chain, but more on this later), but I don't agree that bar-none all tests should use `done()`. Returning a promise as a part of a test is idiomatic mocha, it saves on boilerplate `.then(function() { done(); });` code littering your test suites (and making them harder to read through imho). 

> Please remove chai. It's an extra dependencies that is not required. assert is part of the core, works
> fine and I happen to be familiar with it. I'm not familiar with chai, and since I still maintain the code, it's
> more work for me. An example of why I think chai doesn't do a good job is the description of eql. It says
> "assert that the target is deeply equal to value. It's typically different than assert.deepEqual since it 
> tests for all keys in the object and its prototype chain http://chaijs.com/api/bdd/#eql

I've already agreed to do this, but I'm glad we finally dragged the "I just don't like it" out of you :smile:. In response to the added point there about deep equality, that's another matter of opinion but frankly the I believe chai's take on that to be the "correct" way. When you do [this](https://github.com/neumino/thinky/blob/master/test/model.js#L184) in your tests, you are saying that a plain object `{ id: 'foo' }` is exactly the same, deeply, as the returned Model which is absolutely not the case. If you were to actually compare those two it would fail, and for non-obvious reasons, because the prototypes are different.  I have been bitten by this countless times in many different JS projects, it can lead to massive wastes of time.

> Have the fixture be able to create a model and store whether the table under the hood is ready or not.
> Have this method takes an extra argument to match {init: false}. Model should be recreated, tables
> should not. Tests should take the same time or be faster since we currently recreate tables per 
> "group" of tests (describe) if I'm not mistaken.

You are indeed mistaken. There is only one set of tests that does this currently (the "Models" described section), and while you could certainly rewrite all tests to be that way, it wasn't on my list of things to do. It's a non-trivial amount of work to convert these tests to all run that way, and frankly its generally much easier for a developer to fully create the model in the test they are working on and "just not think about it" (rather than setting `test.models.MyModel`). Again, if rethinkdb's tableCreate took a shorter amount of time (and it will) you wouldn't even notice a difference, but your tests would actually be clean and isolated.

> Do not wait for model to be ready to fire queries. Queries are queued until the model is ready, this is
> by design. It's dev friendly. In prod, it is supposed to be a no op. If it is not, you are probably crazy-ish.
> In prod you are supposed to pass {init: false} and run a dummy script with your models to create the 
> indexes (though I admit I never wrote about that).

This "behind the scenes" work is incomplete. The original reason I peppered the code with calls to `ready()` was because I ran into a few cases (in particular when join tables are automatically created and involved) where the queries didn't in fact wait for that table to be ready, and in turn when the test finished and attempted to clean up and remove the table, it hadn't actually finished creating yet. This is precisely what is wrong with hiding this work behind the scenes, and perhaps we should open a new issue to discuss this in particular. I believe that the default behavior of `createModel` should be to not touch the database at all; as you say the tables are more than likely already created, you're just informing thinky about how it needs to build its queries. 


Anyway, in conclusion you:
  * don't want the preference towards asynchronous behavior
  * don't want chai
  * don't want to clean up and recreate tables between test runs
  * don't want to wait for tables to be ready before test run

additionally, you don't really want the reduction in duplicate code by using for loops but you can live with it.

That basically denies all the points this PR was created to address. I'm fine with that, you're the maintainer, but I can't help but feel that with a slightly more open mind you could dramatically reduce your work load here. 

Cheers

test/model.test.js
mbroadst: @neumino FYI, looks like you'll be forced to fix the deepEqual checking for prototypical equivalence by node itself anyway: https://github.com/nodejs/node/pull/621

neumino: This PR was against io.js and was rejected for breaking too much user-land code. Some people seemed to agree with [my comment](https://github.com/nodejs/node/pull/621#issuecomment-71744908) back there.

mbroadst: This was directly picked from the LTS wiki for breaking changes in nodejs v4.0.0, I think the community has spoken on the issue :)

neumino: Not sure what you are talking about. They added a strict/not strict deep equality. It doesn't go up in the prototype chain as far as I can tell
https://github.com/nodejs/node/pull/639/files

Or maybe you're talking about something else?

mbroadst: https://github.com/nodejs/node/blob/master/lib/assert.js#L194

neumino: This is for assert.deepStrictEqual not assert.deepEqual.

mbroadst: you seem to have missed my correction above

neumino:
```
michel@h9:~/projects/node-stuff/node$ ./node --version
v4.0.0-pre
michel@h9:~/projects/node-stuff/node$ cat test.js
var assert = require('assert');

var a = {};
var b = {};
b.__proto__ = {foo: 'bar'};
try {
  assert.deepEqual(a, b);
  console.log("Assert success.");
}
catch(err) {
  console.log("Assert failure.");
}
michel@h9:~/projects/node-stuff/node$ ./node test.js
Assert success.
```

mbroadst: @neumino if you re-read my comment above I note that the wiki is incorrect, and indeed deepEqual will still be the same with the 4.0.0 release.

PR #340, opened 2015-09-09
Fix Feeds documentation around deleting of data

meenie:
`doc.getOldValue()` is equal to `null` in this instance. Changing it to just `doc` gives you the correct data that was just deleted.


neumino:
Thanks!

PR #344, opened 2015-09-17
Added new custom errors and example

grantcarthew:
I just saw the gh-pages-next branch, so I don't know if this is needed or not?

### DIFF
--- a/documentation/api/thinky.md
+++ b/documentation/api/thinky.md
@@ -90,9 +90,14 @@ var thinky = require('thinky')();
 var Errors = thinky.Errors;
 ```
 
-Custom errors created by `thinky`. For now the only error that thinky will
-create is a `DocumentNotFound`, and is triggered when a query returns `null`
-while it expects to get back a full document.
+Custom errors are thrown by `thinky` to allow your code to respond elegantly.
+These custom errors extend the native `Error` object.
+
+Currently `thinky` will throw the following errors:
+- `DocumentNotFound` is thrown when a `get` call returns `null` instead of a document.
+- `InvalidWrite` is thrown when an in place upgrade/replace returns a non-valid document.
+- `ValidationError` is thrown when the schema validation of a document fails.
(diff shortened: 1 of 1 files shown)


grantcarthew:
Hey @neumino, I have come across an error in my code and it is because I am using Node v0.12 promises mixed with Bluebird promises. As far as I can see, Node promises don't support error types. Is this correct? If this is the case, then maybe the code examples should be moving the error type checks into the catch function rather than relying on Bluebird.

Here is the example I am using:
```
// This function returns a node promise
VerifyAuthToken().then(result => {
  return UserModel.get('invalid id as an example').run()
}).catch(errors.DocumentNotFound, err => {
  // This handler is not being called because node promises don't support types.
}).catch(err => {
 // Unexpected error handler.
  // Should we do type checking here? It would be better supported.
})

```

grantcarthew:
So I just tested this in a single catch block:
```js
    if (err instanceof errors.DocumentNotFound) {
      logger.debug('Bingo')
    }
```
Works fine.

grantcarthew:
Maybe now that node fully supports Promises thinky could move away from Bluebird?

neumino:
Hey @grantcarthew! Thanks for the PR! It looks good. `gh-pages-next` is just an old branch.

Yes natives promises don't support `catch(predicate, handler)`. I don't plan to move to native promises mostly because the native implementation is actually slower than bluebird, so I see very little benefits from doing so. All the features available in native promises are already available in bluebird if I'm not mistaken.

grantcarthew:
Ok, I didn't know that about Bluebird. I will switch to using it myself. Thanks.
I just looked at the page and the bullet points didn't work. Not sure why. It could be the lack of a blank line before the first hypen.

Do you want me to try and fix it? I don't want to cause you more work mate.

neumino:
No worries, I just pushed a fix, it was just missing a line. Thanks for your help!

PR #357, opened 2015-09-25
outdated dependency

feliperoberto:


neumino:
Thanks for following up with a PR :)

PR #360, opened 2015-09-29
Add custom relations tables. Fixes #282

### LINKED ISSUE #282: Custom relation table
Hi !

In 1-n or n-n relationships, is there any way to append custom fields to relation tables ?
In multiple ORMs you can specify `through: otherModel` to specify the custom table to use instead of automatically creating one.

gjuchault:
Adopted syntax :

``` js
var thinky = require('thinky')();

var A = thinky.createModel('A', {id: String});
var B = thinky.createModel('B', {id: String});
var A_B = thinky.createModel('A_B', {id: String, extra: String});
var A_A = thinky.createModel('A_A', {id: String, extra: String});
A.hasAndBelongsToMany(B, 'bs', 'id', 'id', { through: A_B });
B.hasAndBelongsToMany(A, 'as', 'id', 'id', { through: A_B });

A.hasAndBelongsToMany(A, 'as', 'id', 'id', { through: A_A });

A.delete().then(function () {
    return B.delete();
}).then(function () {
    var a = new A({id: 'a1'});
    a.bs = [new B({id: 'b1'})];
    a.as = [new A({id: 'a2'})];

    return a.saveAll({ bs: true, as: true });
}).then(function () {
    // Stash some extra fields on the link.
    return A_B.get('a1_b1').update({extra: '42'});
}).then(function () {
    // Stash some extra fields on the link.
    return A_A.get('a1_a2').update({extra: '42'});
}).then(function () {
    return A.get('a1').getJoin({bs: true, as: true });
}).then(function (r) {
    console.log(r.bs[0]);
    console.log(r.as[0]);
    process.exit(0);
});
```

Missing form this PR : version push, documentation

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1524,6 +1524,11 @@ Document.prototype._deleteHook = function(docToDelete, deleteAll, deletedDocs, d
       var parents = hasOne[key];
       for(var i=0; i<parents.length; i++) {
         delete parents[i].doc[parents[i].key];
+        util.loopKeys(parents[i].doc.__proto__._hasOne, function(joined, joinKey) {
+          if (joined[joinKey].doc === self) {
+            delete parents[i].doc.__proto__._hasOne[joinKey];
+          }
+        })
       }
     });
     util.loopKeys(self.__proto__._parents._belongsTo, function(belongsTo, key) {
@@ -1541,6 +1546,14 @@ Document.prototype._deleteHook = function(docToDelete, deleteAll, deletedDocs, d
       for(var i=0; i<parents.length; i++) {
         for(var j=0; j<parents[i].doc[parents[i].key].length; j++) {
           if (parents[i].doc[parents[i].key][j] === self) {
+            util.loopKeys(parents[i].doc.__proto__._hasMany, function(joined, joinKey) {
+              for(var k=0; k<joined[joinKey].length; k++) {
+                if (joined[joinKey][k].doc === self) {
+                  joined[joinKey].splice(k, 1);
+                  return false;
+                }
+              }
+            });
             parents[i].doc[parents[i].key].splice(j, 1);
             break;
           }
--- a/lib/document.js
+++ b/lib/document.js
@@ -548,12 +548,26 @@ Document.prototype.__makeSavableCopy = function(doc, schema, options, model, r) 
  * @param {boolean} saveAll Whether _save should recurse by default or not
  * @param {Object=} savedModel Models saved in this call
  * @param {Object=} callback to execute
+ * @param {Object=} LinkModel Specify link model if pivot is present
+ * @param {String=} parentId Parent relation id if pivot is present
  * @return {Promise=}
  */
-Document.prototype._save = function(docToSave, saveAll, savedModel, callback) {
+Document.prototype._save = function(docToSave, saveAll, savedModel, callback, LinkModel, parentId) {
   //TOIMPROVE? How should we handle circular references outsides of joined fields? Now we throw with a maximum call stack size exceed
   var self = this;
   self.emit('saving', self);
+
+  var pivotData = undefined;
+  if (LinkModel) {
+    var pivot = self.pivot;
+    delete self.pivot;
+    pivotData = {
+      parentId: parentId,
+      selfId: self.id,
+      data: pivot,
+      model: LinkModel
+    };
+  }
 
   return util.hook({
     preHooks: self._getModel()._pre.save,
@@ -561,7 +575,7 @@ Document.prototype._save = function(docToSave, saveAll, savedModel, callback) {
     doc: self,
     async: true,
     fn: self._saveHook,
-    fnArgs: [docToSave, saveAll, savedModel]
+    fnArgs: [docToSave, saveAll, savedModel, callback, pivotData]
   }).nodeify(callback);
 }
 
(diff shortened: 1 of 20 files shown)


gjuchault:
I don't know why the build is not passing, actually, the link document is not ensured (which is strange : I use promises/callback and saveAll)

gjuchault:
I did update to be able to do that :

```js
var thinky = require('thinky')();

var A = thinky.createModel('A', {id: String});
var B = thinky.createModel('B', {id: String});
var A_B = thinky.createModel('A_B', {id: String, extra: String});
var A_A = thinky.createModel('A_A', {id: String, extra: String});
A.hasAndBelongsToMany(B, 'bs', 'id', 'id', { through: A_B });
B.hasAndBelongsToMany(A, 'as', 'id', 'id', { through: A_B });

A.hasAndBelongsToMany(A, 'as', 'id', 'id', { through: A_A });

A.delete().then(function () {
    return B.delete();
}).then(function () {
    var a = new A({id: 'a1'});

    var b = new B({
        id: 'b1',
        pivot: {
            extra: '42'
        }
    });

    a.bs = [b];
    a.as = [new A({id: 'a2'})];

    return a.saveAll({ bs: true, as: true });
}).then(function () {
    return A.get('a1').getJoin({bs: true, as: true });
}).then(function (r) {
    console.log(r.bs[0]);
    console.log(r.as[0]);
    process.exit(0);
});
```

neumino:
Thanks @gjuchault 

I have friends and family this week at home (yea they thought it would be cool to come at the same time -_-), so I'm pretty busy, but I'll do a quick pass now

neumino:
(and will probably continue this week end)

test/advanced.js
```diff
@@ -436,6 +436,107 @@ describe('Advanced cases', function(){
...
+      var Model = thinky.createModel(modelNames[0], {
+        id: String
+      });
+
+      var OtherModel = thinky.createModel(modelNames[1], {
+        id: String
+      });
+
+      var LinkMOM = thinky.createModel(modelNames[0] + '_' + modelNames[1], {
+        id: String,
```
neumino: You probably need more fields here, I believe we also save the left key and right key in it.
If you set `allowExtra(false)` I believe errors will be throw.

neumino:
So you store the extra information in `pivot`?
Shouldn't you get something like

```
doc = {
  docValues
  links: [
    {linkValues: ..., otherDoc: ...},
    {linkValues: ..., otherDoc: ...},
    {linkValues: ..., otherDoc: ...}
  ]
}
```

lib/document.js
```diff
@@ -631,10 +646,18 @@ Document.prototype._saveHook = function(docToSave, saveAll, savedModel, callback
   var p = new Promise(function(resolve, reject) {
     // Steps:
     // - Save belongsTo
+    // - Save pivot
     // - Save this
     // - Save hasOne, hasMany and hasAndBelongsToMany docs
     // - Save links
 
+    // Save pivot table
+    var firstPromise;
```
neumino: Shouldn't this be saved with the links?

neumino:
Also the build is currently failing because of a bug in rethinkdb (typically if you close a cursor, rethinkdb will return an error)

neumino:
I have to run, I'll try to give it another look tonight or at worst this week end. Thanks @gjuchault

test/advanced.js
gjuchault: Absolutely !

lib/document.js
gjuchault: I did try to save it with links; somehow I couldn't (I don't remember why; but saving the pivot here was working. Maybe we should investigate this).

gjuchault:
Hi,

> So you store the extra information in pivot?
> Shouldn't you get something like ..

In fact, the only way of getting the extra is when you do something like this :
```
query.getAll({ other: true }).then(function (m) {
    m.other.pivot // Here we have the extra from m/pivot
});
```
And I did not try to make something like `model.links` without using getAll.
In fact I did try to inspire from [Laravel's ORM](http://laravel.com/docs/5.0/eloquent#working-with-pivot-tables)

> Also the build is currently failing because of a bug in rethinkdb (typically if you close a cursor, rethinkdb > will return an error)

Okay

neumino:
@gjuchault -- sorry for the late answer. I've commented on the issue
https://github.com/neumino/thinky/issues/282
Let's talk about it there and keep this PR for only technical details :)

Thanks again

gjuchault:
@neumino Okay I'm still stucking on thoses tests. I've added the ternary relationship (not yet pushed); but I've still some failures some times.
I'm kinda desperate actually!

gjuchault:
I'd like to ping again @neumino as this has been discussed in #421. What is the status of this ?

neumino:
Hum, I just pulled the branch. The tests are not passing, so it's not ready as it is.

I looked a bit, and I think it's a bit different than #421. We could solve #421 by saving the direction in the primary key. Now we just alphabetically order them.

pyramation:
Just wanted to bump this back up 👍 this would be great and allow us to use Thinky ;)

gjuchault:
@neumino @MarkHerhold 
Just an update on this

I've come across two problems.
1. I can't find a place in `saveAll` flow where I'm 100% sure the relation has been inserted so that I can update it and add extra fields. I tried adding a `save` hook but it seems that in the test nothing waits for the save hook (hooks are executed in order, but saveAll seems to resolve before) but this make sense. If you look into the last commit I tried to add a `.then` to the `util.hook` here https://github.com/neumino/thinky/pull/360/files#diff-fc7fc7567205f95715daa4204b8ea0b0R579

2. It seems that my query addings are also causing some troubles on some tests because pivot is added on each hasAndBelongsToMany, which causes some `deepEqual` to fail : https://github.com/neumino/thinky/pull/360/files#diff-9c4a8f54117f66844b2eb1d9b8341531L2318
This should be taken care of of course and this modification should be reverted

PR #363, opened 2015-10-09
Expose thinky types directly from module without having to instantiat…

jondubois:
…e thinky


neumino:
Looks good to me.
I'll write a few more stuff tonight for thinky and bundle everything together. It should be out tonight or tomorrow morning.

neumino:
Pushed in 2.1.9.
Thanks for the PR @jondubois

jondubois:
@neumino Thanks!

PR #381, opened 2015-11-04
added duplicate primary key error

no0dles:
allows to catch duplicate primary keys
```
document.save()
	.then(function(r) {
		...
    })
    .catch(Errors.DuplicatePrimaryKey, function () {
        ...
    })
    .error(function (err) {
    	...
    });
```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -809,7 +809,7 @@ Document.prototype._onSaved = function(result, docToSave, saveAll, savedModel, r
   var self = this;
 
   if (result.first_error != null) {
-    return reject(Errors.create(result.first_error));
+    return reject(Errors.create(new Error(result)));
   }
 
   util.tryCatch(function() { // Validate the doc, replace it, and tag it as saved
--- a/lib/document.js
+++ b/lib/document.js
@@ -809,6 +809,12 @@ Document.prototype._onSaved = function(result, docToSave, saveAll, savedModel, r
   var self = this;
 
   if (result.first_error != null) {
+    var duplicatePrimaryKeyRegex = new RegExp('^' + new Errors.DuplicatePrimaryKey().name);
+    if(result.first_error.match(duplicatePrimaryKeyRegex)) {
+      //Reject with an instance of Errors.DuplicatePrimaryKeyError
+      return reject(new Errors.DuplicatePrimaryKey (result.first_error));
+    }
+
     return reject(new Error(result.first_error));
   }
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -809,7 +809,7 @@ Document.prototype._onSaved = function(result, docToSave, saveAll, savedModel, r
   var self = this;
 
   if (result.first_error != null) {
-    return reject(Errors.create(new Error(result)));
+    return reject(Errors.create(new Error(result.first_error)));
   }
 
   util.tryCatch(function() { // Validate the doc, replace it, and tag it as saved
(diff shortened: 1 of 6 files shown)


no0dles:
allows to catch duplicate primary keys
```
document.save()
	.then(function(r) {
		...
    })
    .catch(Errors.DuplicatePrimaryKey, function () {
        ...
    })
    .error(function (err) {
    	...
    });
```

lib/document.js
```diff
@@ -809,6 +809,12 @@ Document.prototype._onSaved = function(result, docToSave, saveAll, savedModel, r
   var self = this;
 
   if (result.first_error != null) {
+    var duplicatePrimaryKeyRegex = new RegExp('^' + new Errors.DuplicatePrimaryKey().name);
```
neumino: Can you move this in errors.js? And add a `Error.create` method that takes a string and return the appropriate error type.
That would let us keep all the logic related to errors in the same places and would make adding new error type easier.
Thanks!

no0dles:
something like this? 

```
module.exports.create = function (message) {
  for(var key in module.exports) {
    if(module.exports[key].prototype instanceof Error) {
      var regex = new RegExp('^' + new module.exports[key]().name);
      if(message.match(regex))
        return new module.exports[key](message);
    }
  }
  return new Error(message);
};
```
I don't know why wrecker gets stuck on the environment setup...

neumino:
No need for a for loop for now. Also keep the regex in a variable, the name may not perfectly match the error message later.

Thanks!

```
DocumentNotFoundRegex = /.../
module.exports.create = function (message) {
  if (message.match(DocumentNotFoundRegex)) {
    return new DocumentNotFoundError(message);
  }
  return new Error(message);
};
```

no0dles:
after a couple of mistakes I think I got what you want

```
module.exports.create = function (error) {
  if(!(error instanceof Error)) {
    return error;
  }

  if(error.message.match(DocumentNotFoundRegex)) {
    return new DocumentNotFound(error.message);
  }

  if(error.message.match(DuplicatePrimaryKeyRegex)) {
    return new DuplicatePrimaryKey(error.message);
  }

  return error;
};
```

lib/errors.js
```diff
@@ -37,3 +39,35 @@ function ValidationError(message) {
...
+    this.message = message;
+};
+DuplicatePrimaryKey.prototype = new Error();
+DuplicatePrimaryKey.prototype.name = "Duplicate primary key";
+var DuplicatePrimaryKeyRegex = new RegExp('^' + DuplicatePrimaryKey.prototype.name);
+module.exports.DuplicatePrimaryKeyRegex = DuplicatePrimaryKeyRegex;
+module.exports.DuplicatePrimaryKey = DuplicatePrimaryKey;
+
+
+module.exports.create = function (error) {
```
neumino: Can you make the function take a string instead?

lib/document.js
```diff
@@ -809,7 +809,7 @@ Document.prototype._onSaved = function(result, docToSave, saveAll, savedModel, r
   var self = this;
 
   if (result.first_error != null) {
-    return reject(new Error(result.first_error));
+    return reject(Errors.create(new Error(result.first_error)));
```
neumino: return reject(Errors.create(result.first_error))

neumino:
(if you are reading your mail, forget the two comments I made).

This looks good to me. I'll merge it later tonight. Thanks @no0dles!

no0dles:
would it be possible to merge this pull request this week?

neumino:
My bad,I forgot about it, doing it now.

neumino:
Hum, I ran into a few issues with bluebird (though I don't think it's related to the PR itself).
I'll take another look at it tonight.

neumino:
Err nevermind, I accidentally bump bluebird to v3. I'll revert that and release it tonight.

PR #382, opened 2015-11-07
add Query.prototype.bind()

ondreian:
I often find that when I am using Thinky with some sort of functional library in a computational chain, one of two things happens.

I end up with a lot of models looking like this:
```js
    id: type.number(),
    age: type.number(),
    admin: type.boolean().default(false)
});

User.defineStatic('bind', function() {
  return this.run.bind(this);
});
```

Or I end up creating some extraneous lines of code:

```js
query = User.find(req.params.id);
boundQuery = query.run.bind(query);
```

It seemed like a useful addition to make thinky play more nicely with functional libraries that handle promises (like Ramda.js for instance).

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -99,12 +99,27 @@ Query.prototype.execute = function(options, callback) {
 }
 
 /**
-* Bind a Query for later use
-* @return {Function} return a bound run
+* Bind Query.prototype.run() for later use
+* @param {Object=} options The options passed to the driver's method `run`
+* @param {Function=} callback
+* @return {Function} return a `this` bound Query.prototype.run()
 */
 
-Query.prototype.bind = function () {
-  return this.run.bind(this);
+Query.prototype.bindRun = function () {
+  var curriedArgs = Array.prototype.slice.call(arguments);
+  return Function.prototype.bind.apply( Query.prototype.run, [ this ].concat( curriedArgs ) );
+}
+
+/**
+ * Bind Query.prototype.execute() for later use
+ * @param {Object=} options The options passed to the driver's method `run`
+ * @param {Function=} callback
+ * @return {Function} return a `this` bound Query.prototype.execute()
+ */
+
+Query.prototype.bindExecute = function () {
+  var curriedArgs = Array.prototype.slice.call(arguments);
+  return Function.prototype.bind.apply( Query.prototype.execute, [ this ].concat( curriedArgs ) );
 }
 
 /**
--- a/lib/query.js
+++ b/lib/query.js
@@ -98,6 +98,14 @@ Query.prototype.execute = function(options, callback) {
   return this._execute(options, false).nodeify(callback);
 }
 
+/**
+* Bind a Query for later use
+* @return {Function} return a bound run
+*/
+
+Query.prototype.bind = function () {
+  return this.run.bind(this);
+}
 
 /**
  * Internal method to execute a query. Called by `run` and `execute`.
--- a/test/query.js
+++ b/test/query.js
@@ -1823,20 +1823,82 @@ describe('In place writes', function() {
     });
   });
 
-describe('Functional Utilities', function () {
-  it('should provide chainable interface to bind runs for later execution', function (done) {
-    var Model = thinky.createModel(modelNames[0], {id: String});
-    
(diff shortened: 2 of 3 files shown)


neumino:
Hum, I'm not super familiar with rambda so I may b missing a use case, but why do you need to bind the promise to the query?
Can you explain your use case?

ondreian:
Sure can!  

Here is a very basic example that illustrates the use case, without bringing in a framework:

```javascript
query = server.models.Metrics.filter({ microservice: req.params.id }).run

pseudoComputationChain = function() { 
  # nested function call like you would have in any functional/control-flow library
  return fn();
};
  
pseudoComputationChain(query).then(function (result) { 
   /* this will never happen
     * run no longer has it's context of an instance of Query
    * throws:
    *
    *   TypeError: this._execute is not a function
    *   at Query.run (C:\dev\npm-modules\thinky\lib\query.js:82:15)
    *
    */
});
```
Another great illustration of this is in the docs for `async` [on binding](https://github.com/caolan/async#binding-a-context-to-an-iterator)

In reality, as opposed to the contrived example in the docs of find a user and seeing if they are admin, I would be querying several instances of databases from a couple different microservices in a job queue to aggregate or run some sort of computation on their results that doesn't fit well in REQL.

neumino:
Oh, I didn't see that you didn't call `run`.

Can you rename your method to `bindAndRun` instead of just `bind`?
`bind` is already used on bluebird promises, and there's still the possibility that we'll add a shortcut to bind a query to a connection.

Also what you want is probably
```
Query.prototype.bindAndRun = function () {
  return this.run()
}
```

ondreian:
I can certainly rename it to something else if you want, I just went with bind due to its similarity with [Function.prototype.bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind?redirectlocale=en-US&redirectslug=JavaScript%2FReference%2FGlobal_Objects%2FFunction%2Fbind), but: 

```javascript
Query.prototype.bindAndRun = function () {
  return this.run()
}
```
Does not work as intended, because you want to return a `Function` bound to the instance of `Query` that calls `Query.prototype.run`, so that when it is called later `this` still resolves to the `Query` instance.

Your example returns a Promise, it is a subtle difference, but very important.

neumino:
My bad, I misread again -_-

In this case can you just rename `bind` to `bindRun`? `boundRun`? I don't have strong opinions on the name as long as it's not just `bind`.

Thanks for the PR!

ondreian:
Will do, I'll probably also add bindExecute, while I am at it, with tests/docs, so both of them are usable in the same way.  Should get these commited tonight or tomorrow morning.

And no problem, thanks for creating such an excellent code base to start with!

ondreian:
@neumino : I improved the implementation a bit of the methods to return bound queries, they now accept curried arguments as well, and also added an improved example in the gh-pages branch. 

The tests were expanded so as to make sure they work the same way for Promises & node-style callbacks. 

It seemed silly in retrospect to only have Promise support, and was an oversight with my personal preference towards them, as the goal was compatibility with as many functional and flow-control libraries as possible and a lot of them still use callbacks.

neumino:
Thanks @ondreian, looks good to me!

paambaati:
@ondreian @neumino This looks super interesting! Can I conditionally chain a `filter` on to another query, say, `getNearest` using this?

Currently, I am doing something like this —

```
let query;
if (!someCondition) {
  query = Model.getNearest(location, queryOptions);
} else {
  query = User.getNearest(location, queryOptions).filter(/* some filter */);
}
return query.run();
```

Is there a nicer way to do this using `.bindRun`, perhaps?

neumino:
@paambaati -- I don't understand your question. Aren't you already conditionally chaining a filter to another query?

paambaati:
I was hoping to do something like this instead, if possible —

```
let query = Model.getNearest(location, queryOptions);
if (someCondition) {
  query = query.filter(/* some filter */);
}
return query.run();

neumino:
You do that, though it has nothing to do with this pull request.

paambaati:
Okay my bad, I got this confused with something else entirely. I'm terribly sorry.

PR #396, opened 2015-11-24
Fix Errors prototype chain

vieks:
You forgot to overwrite the default prototype.constructor property with the new Error classes created.
I also replace the prototype from
```javascript
new Error
```
with
```javascript
Object.create(Error.prototype)
```
Which is a cleaner way to do an inheritance (not affecting 'this' properties from Error to proto chain ).

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -6,7 +6,8 @@ function DocumentNotFound(message) {
   Error.captureStackTrace(this, DocumentNotFound);
   this.message = message || "The query did not find a document and returned null.";
 };
-DocumentNotFound.prototype = new Error();
+DocumentNotFound.prototype = Object.create(Error.prototype);
+DocumentNotFound.prototype.constructor = DocumentNotFound;
 DocumentNotFound.prototype.name = "Document not found";
 module.exports.DocumentNotFound = DocumentNotFound;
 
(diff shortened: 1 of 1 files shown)


vieks:
Update: I commit again, I was too fast while copy/paste my change from DocumentNotFound to the other classes -_-

neumino:
Hum, I'm not sure what this pull request does. `constructor` is defined if you use `new Error`. If you use `Object.create` you also don't have the field `message`.

vieks:
IMHO yes !
> constructor is defined if you use new Error

Each prototype must have a 'constructor' property referencing the function constructor which has provided the prototype.
With your solution:
```javascript
DocumentNotFound.prototype = new Error;
```
DocumentNotFound.prototype. havn't 'constructor' property !
But If you do:
```javascript 
DocumentNotFound.prototype.constructor // == Error
```
**It have one ??? NO !!!**
It's the 'constructor' property inherited from prototype chain, here from Error.prototype that's why it's a reference to Error not DocumentNotFound ! So this is a lack of... We need to define constructor property because:
1) The constructor property inherited reference the wrong class
2) It break the convention specifying each prototype object must have it's **own** constructor prop

Since every function have a 'name' property (ES6) with a proper inheritance you can do:
```javascript
DocumentNotFound.prototype = Object.create(Error.prototype);
DocumentNotFound.prototype.constructor = DocumentNotFound;

...

const err = new DocumentNotFound;
console.log(err.constructor.name); // => "DocumentNotFound"
```
Otherwise you can also define error name directly on function constructor, it's a common pattern for Errors:
```javascript
function DocumentNotFound(msg){
    this.name = 'DocumentNotFound';
    this.message = ...
}
```

> If you use Object.create you also don't have the field message

You define it in the constructor already:
```javascript 
function DocumentNotFound(message) {
   Error.captureStackTrace(this, DocumentNotFound);
   this.message = message || "The query did not find a document and returned null.";
 };
```

neumino:
Fair enough, works for me.

Issue #417, opened 2016-01-05
error refactoring


mbroadst:
I took an initial stab at error refactoring here in #231 which sort of fell by the wayside, but there [has](https://github.com/neumino/thinky/issues/352) [been](https://github.com/neumino/thinky/issues/395) [sufficient](https://github.com/neumino/thinky/issues/395) [discussion](https://github.com/neumino/thinky/issues/328) to warrant a discussion of refactoring.

The first step I took (in #231) was to unify all Errors as subclasses of a base `ThinkyError`, the benefit being mainly one of error consistency, as well as being able to catch all thinky-specific errors (currently all thinky errors are just "inherit" Error and I can't provide a catchall).  This would also open the door for implementing custom `ThinkyError`s for use with e.g. mixins.

Secondly, a point was brought up about not using `"use strict"` in the file, which I believe was properly refuted by my comments there. I would love an explanation for why not to use strict mode, maybe I've been making a mistake my whole time using node.

I think these changes set the stage for the improvements outlined in #328, to make validation errors more programmatically accessible (my guess is that this is the primary use case for them). I'll let discussion on those continue over there, and maintain this issue simply as a discussion for refactoring existing errors.

Thanks!

neumino:
Ok, let's make this issue only about error refactoring. I thought that RethinkDB would bring error code, but it has been pretty long and I haven't heard about that so let's start with Thinky.

On top of my head, here's what I can think of
- validation error
- document not found
- invalid write
- duplicate primary key
- programmer error (passing a negative length for an array for example).

1. The main pain from my understanding is around validation error right? Because they just give a string that you have to run against a regex. So let's have the following fields:
- `message`: The same as what we have now
- `path`: an array to the field that did throw. For example for
```
{
  foo: {
    bar: [ {buzz: 2}, {buzz: 3}]
  }
}
// The paths could be:
  - ['foo']
  - ['foo', 'bar']
  - ['foo', 'bar', 0]
  - ['foo', 'bar', 0, 'buzz']
  - ['foo', 'bar', 1, 'buzz']
```
- `type`: Not exactly sure what to put here, The error could be: 
  * field missing
  * field null
  * wrong type
  * custom validator failed

For document not found we could add the primary key that did not return in an extra field. So we would have `message` and `primaryKey`
Same for duplicate primary key

I think invalid write errors are fine the way they are.


[...]

PR #420, opened 2016-01-07
refactor(errors): provide a proper error hierarchy

mbroadst:
This is the first change based on the discussion in #417, bringing the errors in thinky into a strict hierarchy based on a common `ThinkyError` base.

This also includes the change to provide the `primaryKey` in the `DocumentNotFound` error, as well as bump to use semver for `rethinkdbdash` to alleviate wercker errors.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,80 +1,84 @@
-var DOCUMENT_NOT_FOUND_REGEX = new RegExp('^The query did not find a document and returned null.*');
-module.exports.DOCUMENT_NOT_FOUND_REGEX = DOCUMENT_NOT_FOUND_REGEX;
-
-var DUPLICATE_PRIMARY_KEY_REGEX = new RegExp('^Duplicate primary key.*');
-module.exports.DUPLICATE_PRIMARY_KEY_REGEX = DUPLICATE_PRIMARY_KEY_REGEX;
-
+"use strict";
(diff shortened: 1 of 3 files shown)


neumino:
Why not just create a new function ThinkyError, and have all the current errors'prototype be an instance of ThinkyError instead of Error?

mbroadst:
@neumino that's exactly what's happening here, except that `ThinkyError` inherits from `Error` as well. Are you suggesting that `Error` not be in the chain at all? I think the better question would be why would you _not_ do that, as it is otherwise conventional JavaScript (these _are_ in fact errors).

neumino:
I mean, why not just replacing
```
-DocumentNotFound.prototype = Object.create(Error.prototype);		
```
With
```
-DocumentNotFound.prototype = Object.create(ThinkyError.prototype);		
```

We just want the inheritance for the prototype chain right? We actually don't need anything else?

mbroadst:
* I felt that the existing code was inconsistent (using `Object.create`, or just `new Error()`) 
* the resulting code is more of an idiomatic node-style (using built-in methods primarily)
* the resulting code ends up in less lines, and is more readable imho

neumino:
Ouki, merged and released in 2.2.4

neumino:
Thanks!

PR #459, opened 2016-03-13
update bluebird version in order to prevent starvation by nextTick

ramimoshe:
In nodejs 4.* it is not recommend to use .nextTick() because it can cause to starvation, 
and the recommend function is setImmediate() 

the older bluebird version used nextTick()

https://github.com/nodejs/node/wiki/API-changes-between-v0.10-and-v4

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies":{
     "rethinkdbdash": "^2.2.15",
-    "bluebird": "~2.1.3",
+    "bluebird": "~2.10.2",
     "validator": "~ 3.22.1"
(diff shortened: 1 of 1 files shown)


neumino:
Same comment as in #430, this breaks the current API.

ramimoshe:
but i upgrade to latest bluebird version 2.* (from 2.1.3 to 2.10.2)
and i run "npm test" before i pushed it
so it should be ok

neumino:
My bad, I thought you were bumping to 3.x.
Merging, thanks : )

PR #462, opened 2016-03-16
Support joins with allowExtra and/or removeExtra

### LINKED ISSUE #455: allowExtra and removeExtra overwrite model joins
I have two models:

```
var ThingA = thinky.createModel('ThingA', type.object().schema({
  id: type.string()
}).allowExtra(false));

var ThingB = thinky.createModel('ThingB', type.object().schema({
  id: type.string(),
  thingAid: type.string().required()
}).allowExtra(false));

ThingA.hasMany(ThingB, 'stuff', 'id', 'thingAid');
```

If I try to do ThingA.get('someid').getJoin().run(), then I get:
```Extra field `stuff` not allowed```

Similarly, if I use removeExtra() instead of allowExtra(false), then the field 'stuff' appears to be removed from the join.

rosskukulinski:
Fixes #455 and includes two tests to validate the functionality with hasMany join.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -128,6 +128,7 @@ Model.new = function(name, schema, options, thinky) {
         }
       }
     });
+    doc._getModel()._schema._setModel(doc._getModel());
     if (proto.needToGenerateFields === true) {
       doc._generateDefault();
     }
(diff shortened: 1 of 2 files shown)


rosskukulinski:
Hmm - build failed.  Not sure why the git checkout failed in wercker

neumino:
Nice, good catch @rosskukulinski!

neumino:
Wercker has some ressouces issues sometimes, but I manually tested it, looks good.

PR #468, opened 2016-03-29
Added string uuid validation.

artdude543:
Added UUID validation for easier validation when working with UUIDs on models.

- I've opted to support versions 3, 4 and 5 due to the validation module used only supports these.
- I'll send a PR to the docs branch also when if this is pulled in.

### DIFF
--- a/lib/type/string.js
+++ b/lib/type/string.js
@@ -262,6 +262,9 @@ TypeString.prototype.default = function(fnOrValue) {
  * @return {TypeString}
  */
 TypeString.prototype.uuid = function(version) {
+  if (isNaN(version)) {
+    throw new Errors.ValidationError("The value for `version` must be a number.");
+  }
   if (version < 3 || version > 5) {
     throw new Errors.ValidationError("The value for `version` must be either 3, 4 or 5");
   }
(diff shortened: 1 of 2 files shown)


artdude543:
Any chance of this being looked at? The failed test does not have anything to do with this PR (If that's why it's not been looked at)

neumino:
Arg sorry I missed that. I'll look into it tonight.

lib/type/string.js
```diff
@@ -251,6 +256,19 @@ TypeString.prototype.default = function(fnOrValue) {
   return this;
 }
 
+/**
+ * Set the string to be a uuid.
+ * @param {number} version
+ * @return {TypeString}
+ */
+TypeString.prototype.uuid = function(version) {
+  if (version < 3 || version > 5) {
```
neumino: Can you also check that version is a number?

neumino:
Besides the one little comment, that looks good to me.

artdude543:
I'll add the number check now. Not sure how I actually forgot to add that in the first place xD

artdude543:
Huh, well the tests passed locally O_o gonna check the code again.

artdude543:
I've removed the test for number checking as I'm not sure how to create a test which validates the model. As the error would be thrown from the Model and I'm new to tests so still learning :/

neumino:
Thanks, merged!

neumino:
Can you follow up with a PR on the `gh-pages` branch? Thanks!

PR #490, opened 2016-04-28
Accept table options when creating a model

AndrewBarba:
Expose passing options to internal `r.tableCreate` which allows us to do things like set the table durability, replicas, etc.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -43,6 +43,9 @@ function Model(name, schema, options, thinky) {
 
   this._pk = (options.pk != null) ? options.pk : 'id';
 
+  this._table = (options.table != null) ? options.table : {};
+  this._table.primaryKey = this._pk;
+
   this._thinky = thinky;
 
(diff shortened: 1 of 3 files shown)


AndrewBarba:
These tests should not have failed... I ran them a dozen times locally. Can you re-run from wercker?

neumino:
Looks good, thanks.

Yea, wercker has been annoying lately, it looks like it has a hard time starting rethinkdb and running tests...

AndrewBarba:
Have you tried using the docker environment on Wercker? That's what I use and it's is quite reliable. Here is the top of my `wercker.yml`:

```
box: node:4.4.3

services:
  - rethinkdb:2.3.1
```

And then to connect:

```
rethinkdb: {
  host: process.env.RETHINKDB_PORT_29015_TCP_ADDR || 'localhost'
}
```

neumino:
No I haven't, but maybe I should : )

Issue #86, opened 2014-05-21
Support for embeddable objects


tlvenn:
Hi Neumino,

I am playing with thinky and when it comes to relationship, the assumption is to leverage the join feature of rethinkdb which is great. However in some scenarios, it's preferable to embed rather than to join.

To some degrees, thinky might already support this with nested objects but it forces you to inline all your embeddables and I am not so sure it's ideal to support collection of elements.

Maybe if we could declare our embeddables with thinky.createEmbeddable and then reference them in the schema where they are embedded.

Now i might be biased by my prior JPA experience and maybe this whole thing is unnecessary when it comes to json documents...

What do you think ?

cur3n4:
+1

I think the ability to define virtual fields which operate just on the nested model properties would be great. e.g:

```javascript
var User = thinky.createModel("User", {
  id: String,
  contact: {
    email: String,
    phone: String,
    fullContact: type.virtual().default(function() {
        return this.email+" "+this.phone;
    }
  }
});
```

neumino:
@cur3n4 -- you can already do that.

cur3n4:
@neumino Indeed! However it doesn't seem to work for arrays of nested objects. I get a TypeError: Cannot read property '0' of undefined.

Ideally defining methods on the nested model (not only virtual) would be very useful.

neumino:
@cur3n4 -- can you open a new issue with the full backtrace and a snippet to reproduce this error? Thanks

cur3n4:
@neumino Done: [#375](https://github.com/neumino/thinky/issues/375)

cur3n4:
From my point of view, the main thing I am missing about nested models is the ability to define methods (not virtual properties). Something like this:

```javascript
var User = thinky.createModel("User", {
  id: String,
  contact: {
    email: String,
    someMethod: type.method().default(function(someParameter1, someParameter2) {
        // More complex logic than this...
        return this.email == someParameter1 || this.email == someParameter2;
    }
  }
});
```

PR #497, opened 2016-05-09
Support for functions on sub-docs

davefinster:
Keen for some feedback on some small additions I made that allow for the attachment of functions to sub-docs. I was porting a codebase from MongoDB to RethinkDB and this functionality was required.

Sub-doc functions can be defined by simply adding then as a member of the object that would otherwise define the schema (as shown in the tests). When a document is instantiated, the '_overrideProperties' method is invoked to define getters/setters for any sub-doc properties such that the prototype can be injected. It does something similar for arrays of sub-docs by hijacking the push method on the attached array.

To support this, I've had to add the '_shadowObject' property to the document, which does mess with validation (which I've accounted for by adding exceptions in validation and ensuring it doesn't make it into the saveableCopy) and ideally I'd like to be able to hide it - but not sure what the best way of doing that is.

### DIFF
--- a/test/sub_doc_functions.js
+++ b/test/sub_doc_functions.js
@@ -1,0 +1,171 @@
+var config = require(__dirname+'/../config.js');
+
+var thinky = require(__dirname+'/../lib/thinky.js')(config);
+var r = thinky.r;
+var type = thinky.type;
+
+var util = require(__dirname+'/util.js');
+var assert = require('assert');
+var Promise = require('bluebird');
+var Errors = thinky.Errors;
+
+
+var modelNameSet = {};
+modelNameSet[util.s8()] = true;
+modelNameSet[util.s8()] = true;
+
+var modelNames = Object.keys(modelNameSet);
+
+var cleanTables = function(done) {
+  var promises = [];
+  var name;
+  for(var name in modelNameSet) {
+    promises.push(r.table(name).delete().run());
+  }
+  Promise.settle(promises).error(function () {/*ignore*/}).finally(function() {
+    // Add the links table
+    for(var model in thinky.models) {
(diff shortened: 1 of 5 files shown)


test/sub_doc_functions.js
```diff
@@ -0,0 +1,175 @@
...
+  });
+
+  var basicModelThreeSpec = {
+    id: String,
+    name: String,
+    email: type.string(),
+    address: {
+      street: String,
+      geoCode: function() {
+        return this.street;
```
neumino: So what's the difference with `define`? This doesn't seem to allow things that are currently not allowed right?

davefinster: The difference is that (from my understanding of the doco), `define` can only add functions to the top level model object, not sub-docs/sub-objects.

In the case of `basicModelThreeSpec `, `address` is a sub-object of the model that owns the function `geocode`, rather than the function having to be defined onto the model itself.

The changes also apply to arrays of sub-objects by allowing each object to have functions added to them based on the schema definition

neumino: Hum, that's still adding a decent chunk of complexity without providing a new functionality, so I'm not sure it's worth the maintenance effort.

davefinster: Maybe I'm not explaining it well enough. The changes help address the query at the end of #86 where cur3n4 asks for this:

```javascript
  id: String,
  contact: {
    email: String,
    someMethod: type.method().default(function(someParameter1, someParameter2) {
        // More complex logic than this...
        return this.email == someParameter1 || this.email == someParameter2;
    }
  }
});`

Aside from defining virtual properties, there is (AFAIK) no way to define a method or function on a sub-document of a model?

PR #524, opened 2016-07-04
Allow negative range bounds for number types

simonratner:


neumino:
Thanks @simonratner! -- I have no idea why we were enforcing positive numbers :/

PR #555, opened 2016-08-31
Add pingInterval option to config

### LINKED ISSUE #554: Add pingInterval option to config
The server is closing connection after a time, which makes our staging environment unavailable. We don't want to restart the server several times because of this issue (by exiting with non-zero error code), but I found the solution on the official rethinkdbdash.

    The connection was closed by the other party.

When importing the driver, I can set

- `pingInterval`: - if `> 0`, the connection will be pinged every `pingInterval` seconds, default `-1`

https://github.com/neumino/rethinkdbdash

It was implemented following this discussion, which describes the issue pretty well:
https://github.com/neumino/rethinkdbdash/issues/192#issuecomment-212753388

Can we add this option to the config, so it will ping the server every `x` seconds and avoid the connections getting closed?

balintant:
The server is closing connection after a time, which makes can cause
errors in several environments.

This commit adds a new config option when initializing thinky, which
will help keeping the connection alive, using `rethinkdbdash`'s solution
with the same input.

- `pingInterval`: - if `> 0`, the connection will be pinged every
  `pingInterval` seconds, default `-1`
  https://github.com/neumino/rethinkdbdash

It was implemented following this discussion, which describes the issue
pretty well:
https://github.com/neumino/rethinkdbdash/issues/192#issuecomment-212753388

### DIFF
--- a/lib/thinky.js
+++ b/lib/thinky.js
@@ -44,6 +44,10 @@ function Thinky(config) {
   self._options.validate =
     (config.validate != null) ? config.validate : 'onsave';
 
+  // The connection will be pinged every <pingInterval> seconds
+  self._options.pingInterval =
+    (config.pingInterval != null) ? config.pingInterval : -1;
+
   if (config.r === undefined) {
     self.r = rethinkdbdash(config);
   }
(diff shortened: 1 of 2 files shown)


balintant:
Just found out that config is passed to `rethinkdbdash` directly, which means no need for this change.

PR #558, opened 2016-09-10
Add _rel option to getJoin

tslocke:
Say we have a forum with posts:

```
Forum.get(forumId).getJoin({posts: true})

// Only recent posts
Forum.get(forumId).getJoin({
  posts: {_apply: posts => posts.orderBy(r.desc('createdAt').limit(10)}
})
```

What if we want recent posts _and_ pinned posts?. The `_rel` option lets 
you specify the name of the relation, so it can differ from the the attribute to be merged.

```
Forum.get(forumId).getJoin({
  posts: {_apply: posts => posts.orderBy(r.desc('createdAt').limit(10)},
  pinnedPosts {_rel: 'posts', _apply: posts => posts.filter(p => p('pinned'))}
})
```

Another example - fetch a count along with the posts:

```
Forum.get(forumId).getJoin({
  posts: {_apply: p => p.orderBy(r.desc('createdAt').limit(10)},
  postsCount: {_rel: 'posts', _apply: p => p.count(), _array: false} 
})
```

If you like the feature but there is something lacking in the PR, please let me know and I'll work
on it.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -308,8 +308,8 @@ Query.prototype._convertGroupedData = function(data) {
   }
 }
 
-function extendQueryWithJoin(doc, model, joinName, join, getAll, r, ifPresent, ifMissing) {
-  var innerQuery;
+function queryForJoin(doc, model, joinName, join, ifPresent, ifMissing) {
+  var innerQuery, r = join.model._thinky.r
 
   switch (join.type) {
     case 'hasOne':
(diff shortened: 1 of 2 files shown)


MarkHerhold:
Why would you not just do two queries? Is it just to save one call to the DB?

tslocke:
That's what getJoin is for in the first place, right? The performance gain
is very worthwhile.

One of the things I love about REQL is that you can easily get everything
you need in one request.

MarkHerhold:
@tslocke I was just asking. :)

tslocke:
@neumino I have a bunch more enhancements to getJoin which I hope you will like. Need to decide on this one first though.

tslocke:
@neumino any interest in this PR?

neumino:
Sorry I missed this PR.

I think you can do everything you describe with `fold`. Also you can define multiple relations with the same table and use different compound indexes to make things more efficient.

Or did I miss a use case?

tslocke:
I think the question should be more like, does the convenience justify the small addition to the API? 

After all, everything Thinky does can be done directly with the db driver.

That said, unfortunately I'm moving my projects off rethinkdb since the bad news, so I'm not going to push the point. Please go ahead and close the PR if you wish.

I also added a scopes like functionality to Thinky, on top of this enhancement (like scopes in Sequelize or ActiveRecord) which I found extremely useful. If you want to take a look at the fork in my github account please do - I'll send a PR if you want one but I won't do it preemptively.

PR #559, opened 2016-09-10
Fixing hasAndBelongsToMany shard/replica configuration

MarkHerhold:
This PR fixes an issue Tables created by adding a `hasAndBelongsToMany` relation do not properly "inherit" the table configuration from the parent model.

For example, when running a production RethinkDB instance, I configure all the models (and therefore the tables) to have 3 replicas. Tables that are created as a result of the `hasAndBelongsToMany` relation always have 1 replica and 1 shard.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -558,7 +558,9 @@ Model.prototype.hasAndBelongsToMany = function(joinedModel, fieldDoc, leftKey, r
 
   var linkModel;
   if (thinky.models[link] === undefined) {
-    linkModel = thinky.createModel(link, {}, { table: self._table }); // Create a model, claim the namespace and create the table
+    // Create a model, claim the namespace and create the table
+    // passes table options to the underlying model (e.g. replicas, shards)
+    linkModel = thinky.createModel(link, {}, { table: options.table });
   }
   else {
     linkModel = thinky.models[link];
--- a/lib/model.js
+++ b/lib/model.js
@@ -558,7 +558,7 @@ Model.prototype.hasAndBelongsToMany = function(joinedModel, fieldDoc, leftKey, r
 
   var linkModel;
   if (thinky.models[link] === undefined) {
-    linkModel = thinky.createModel(link, {}); // Create a model, claim the namespace and create the table
+    linkModel = thinky.createModel(link, {}, { table: self._table }); // Create a model, claim the namespace and create the table
   }
   else {
     linkModel = thinky.models[link];


MarkHerhold:
I should also note that I am not sure how to test that a table configuration is properly passed in. If you can provide a hint I'd be happy to try and test this.

MarkHerhold:
Ping @neumino

neumino:
Isn't what you want to take some extra options on `hasAndBelogsToMany` and use it to create the link table?

lib/model.js
```diff
@@ -558,7 +558,7 @@ Model.prototype.hasAndBelongsToMany = function(joinedModel, fieldDoc, leftKey, r
 
   var linkModel;
   if (thinky.models[link] === undefined) {
-    linkModel = thinky.createModel(link, {}); // Create a model, claim the namespace and create the table
+    linkModel = thinky.createModel(link, {}, { table: self._table }); // Create a model, claim the namespace and create the table
```
MarkHerhold: `self._table` is simply the table settings (e.g. number of replicas and shards) object for the current model. I'm just passing the table settings along to the model that will inevitably create the n-n relation.

MarkHerhold:
Personally, I would rather the hasAndBelogsToMany relation just assume that the table settings are the same for both the original model as well as the model created for the n-n relation, but I could see how that may not be ideal is some odd use cases.

MarkHerhold:
@neumino Thoughts on this? Do you want the PR to instead add table options that can be passed to hasAndBelongsToMany and defaults to the parent model's table options? I think this is slightly more confusing, but if that is what you want to see, I can look into it.

neumino:
Sorry I missed that.

The problem with the current proposal is for this situation:
- The table Post has 1 shard
- The table Comment has 3 shards

```
Post.hasAndBelongsToMany(Comment, ...)
Comment.hasAndBelongsToMany(Post, ...)
```

Would be different than
```
Comment.hasAndBelongsToMany(Post, ...)
Post.hasAndBelongsToMany(Comment, ...)
```

Can we just add options to hasAndBelongsToMany to take the shard/replica settings?

MarkHerhold:
@neumino Good point. Updated PR.
I'll add docs for the `options.table` field in another PR if this approach is acceptable.

neumino:
Looks good to me, can you follow up with a PR on gh-pages? Thanks!

MarkHerhold:
@neumino will do, expect it today

MarkHerhold:
https://github.com/neumino/thinky/pull/572

MarkHerhold:
@neumino Can you please publish a patch release with these changes?

neumino:
Sorry forgot to do that. It's out in 2.3.7

MarkHerhold:
@neumino Thank you!

PR #597, opened 2016-11-30
Schemas Documentation Updates

LoganArnett:
Updated the documentation for Schemas as I saw that there were inconsistencies and duplicate code examples

It appears that declaring 
```
var type = thinky.type;
```
is covered in the quickstart and can possibly be removed here instead of having variations that include it while others that do not.

### DIFF
--- a/documentation/schemas.md
+++ b/documentation/schemas.md
@@ -122,6 +122,7 @@ This is the only guarantee. A default value should not rely on another default v
 _Example_: Create a basic Model for a `user`.
 
 ```js
+var type = thinky.type;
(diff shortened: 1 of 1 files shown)


neumino:
I'm not sure we should expect people to read the quickstart before reading the doc. These two pages are orthogonal.

LoganArnett:
Then should each example include `var type = thinky.type`?

neumino:
Yep.

LoganArnett:
Updated to be consistent with `var type` in all of the schema examples

neumino:
Sounds good, thanks @LoganArnett

PR #603, opened 2017-01-07
rename event `delete` to `deleted` in documentation

isaackwan:
Took me a while to figure this out.

[Referencing test](https://github.com/isaackwan/thinky/blob/master/test/event.js#L66)

Isaac


neumino:
Thanks!

